Issue #16, opened 2020-10-12
Transcripts


tomrossi7:
1. I would suggest that `podcast:transcript` and `podcast:captions` are really the same thing provided in different formats.  [The way we have approached it with Buzzsprout](https://rss-extension.buzzsprout.com/1.0/podcast/) makes use of the XML `type`. I know this may go against your Goal #2, but it really does accurately capture what is being represented and avoids creating a new tag when people want to make use of another format for transcripts e.g. JSON, WebVTT.

2. Transcript language seems redundant with the language of the podcast which may be better captured with `podcast:language`.

theDanielJLewis:
1. But the idea of captions is for time-based transcripts. I can see uses for both fields. Look at how Apple TV supports "What did they say?" It jumps back a few seconds and turns on captions for a few seconds. That kind of time-based transcript could be helpful even for those without impaired hearing. This requires an extra layer of data beyond transcripts, but maybe that layer can be included inside transcripts. We would need to make captions fully SRT-compatible.
2. While this sounds good and could be the assumption, there could be cases where someone provides a translated transcript of their podcast episode. Then, that would need a separate transcript tag and separate language.

daveajones:
@tomrossi7 I looked at the link you provided and I see your meaning.  I think it comes down to name confusion with the tag.  Are you saying that the aggregators would know this element refers to captions by looking for type="application/srt"?  That works but feels semantically strange to take intent from a mime type.  Is there a way to firm that up with something like the following?

`<podcast:transcript type="application/srt" rel="captions">https://host.com/1.srt</podcast:transcript>`

I agree with @theDanielJLewis on point 2.  I heard a podcast yesterday (a random one one castcoverage) and they started the show saying that this was their first english language episode.  That show would really need the ability for separating the language of the show from the language of the transcript.

tomrossi7:
Great idea on point #2! We don't support transcripts in other languages at this point, but that is a great idea for the future!

I believe all transcripts are time-based, the only thing that changes is the "fidelity". Even our HTML transcripts have timestamps automatically inserted based on monologues. I just don't know if warrants another tag when really it points to the exact same resource represented in another format that is captured in the `type`.

@daveajones what would other `rel` options be for a transcript other than `captions`? Buzzsprout currently serves up feeds in HTML, SRT, and JSON. So is this the proposed RSS standard?

```
<podcast:transcript type="application/srt"  rel="captions">https://host.com/1.srt</podcast:transcript>
[...]

PR #19, opened 2020-10-12
Transcript Updates?

tomrossi7:
These are the changes we've been discussing in issue #16 . What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,22 +121,11 @@ this larger namespace.
 
    Item
 
-   (optional | single)
+   (optional | multiple)
 
-   Links to an external file containing a transcript.  The mime type of the file should be given - such as text/plain, text/html, etc.
+   Links to an external file containing a transcript.  The mime type of the file should be given - such as `text/plain`, `text/html`, `application/srt`, `application/json`.
 
-   All attributes are required.
-
-
-- **\<podcast:captions language="[language code]" type="text/srt">**[url to a SRT captions file]**\</podcast:captions>**
-
-   Item
-
-   (optional | single)
-
-   Links to an industry standard closed-caption/subtitle file format.
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -121,22 +121,11 @@ this larger namespace.
...
-   Item
-
-   (optional | single)
+   (optional | multiple)
 
-   Links to an industry standard closed-caption/subtitle file format.
+   Links to an external file containing a transcript.  The mime type of the file should be given - such as `text/plain`, `text/html`, `application/srt`, `application/json`.
 
-   All attributes are required.
+   The language of the transcript can be overwritten by providing the optional language attribute. All other attributes are required.
```
tomrossi7, line 128: What about making the language optional? In most cases we have 3 transcript types available for each episode in the RSS feed. This would reduce the amount of code required as long as we are comfortable defaulting to the same language as the podcast.

README.md
```diff
@@ -121,22 +121,11 @@ this larger namespace.
...
-
-   All attributes are required.
-
-
-- **\<podcast:captions language="[language code]" type="text/srt">**[url to a SRT captions file]**\</podcast:captions>**
-
-   Item
-
-   (optional | single)
+   (optional | multiple)
```
tomrossi7, line 124: This should be multiple if the transcripts is available in different formats.

tomrossi7 (commented):
@daveajones what do you think?

README.md
daveajones: That's fair.  Let's do that.  I may clarify your statement here as we go refine the wording of things.

tomrossi7: Sure thing! I'm happy to help, but am more comfortable with code than english 🤣

PR #22, opened 2020-10-13
Added Podchaser to serviceslugs.txt

bslinger:
Podchaser can be linked to based on iTunes IDs and RSS feed URLs currently, but of course the direct ID will always be more efficient and accurate.


jamescridland:
👍

daveajones:
Thanks Ben!

PR #26, opened 2020-10-13
Introducing xmlns 1.0 document 

tomrossi7:
What do you think of this format? As tags are agreed upon, this could be the document that captures the work. Additionally, it would serve as the xmlns document to be linked from conforming RSS feeds.

### DIFF
--- a/docs/1.0.md
+++ b/docs/1.0.md
@@ -13,10 +13,13 @@ This tag is used to link to a transcript or closed captions file. Multiple tags 
 
 #### Attributes
 **url:** URL of the podcast transcript
+
 **type:** Mime type of the file such
 as `text/plain`, `text/html`, `application/srt`, `application/json`
+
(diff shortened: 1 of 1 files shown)


docs/1.0.md
```diff
@@ -0,0 +1,30 @@
...
+#### Attributes
+**url:** URL of the podcast transcript
+
+**type:** Mime type of the file such
+as `text/plain`, `text/html`, `application/srt`, `application/json`
+
+**language (optional):** The language of the linked transcript. If there is no language attribute given, the linked file is assumed to be the same language that is
+specified by the RSS `<language>` element.
+
+**rel (optional):** If the rel="captions" attribute is present, the linked file is considered to be a closed captions file, regardless of what the mime type is.  In that scenario, time codes are assumed to be present in the file in some capacity.
```
tomrossi7, line 23: I'm still not totally clear on the use of the `rel` tag. The way it reads, I'm not sure if you should or should not provide it when you use a mime type of something like `srt`.

PR #28, opened 2020-10-13
Podcast Chapter Proposal

normand1:
Podcast Chapters are a feature supported by many [podcast clients](https://stratechery.com/outline/what-podcast-player-do-you-recommend/) already however the current standard is to [embed the chapter content into the MP3 file itself](https://id3.org/id3v2-chapters-1.0). This is podcast specific content and should be moved to the podcast RSS feed directly so it can be more easily accessed and updated by podcasters.

There is some precedent for moving this into the RSS Feed. The [Simple Chapters Specification](https://podlove.org/simple-chapters/) was introduced in 2012 and I've drawn on that spec heavily for this proposal for introduction in the Podcast Namespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,21 @@ this larger namespace.  But, we don't want to be so general that the spec become
    specified by the RSS \<language\> element.  If the rel="captions" attribute is present, the linked file is considered to be a closed captions file, regardless of what the
    mime type is.  In that scenario, time codes are assumed to be present in the file in some capacity.
 
+- **\<podcast:chapters/>**
+
+   Channel (optional | single)
+
+   This element lists optional `<podcast:chapter/>` item elements.
+
+- **\<podcast:chapter start="[(Normal Play Time)]" title="[(string)]" href="[(uri to content)]" type="[mime type]"/>**
+
+   Item (optional | multiple)
+
+   This element specifies a point in time during the podcast that can be linked to directly and optionally supplemented with additional content.
+   `start` is required and uses the [Normal Play Time](https://www.w3.org/TR/media-frags/#naming-time) standard.
+   `title` is required and is used as a user facing identifier for this chapter
+   `href` is optional and points to additional user facing content
+   `type` is optional but it is strongly encouraged if `href` is used so that clients can handle the chapter content appropriately
 
 - **\<podcast:alternateEnclosure type="[mime type]" length="[(int)]" bitrate="[(float)]" title="[(string)]" stream>**[uri of media asset]**\</podcast:alternateEnclosure>**
 


tomrossi7:
I like the idea of bringing chapters into the RSS feed!

README.md
```diff
@@ -116,6 +116,21 @@ this larger namespace.  But, we don't want to be so general that the spec become
...
+   Channel (optional | single)
+
+   This element lists optional `<podcast:chapter/>` item elements.
+
+- **\<podcast:chapter start="[(Normal Play Time)]" title="[(string)]" href="[(uri to content)]" type="[mime type]"/>**
+
+   Item (optional | multiple)
+
+   This element specifies a point in time during the podcast that can be linked to directly and optionally supplemented with additional content.
+   `start` is required and uses the [Normal Play Time](https://www.w3.org/TR/media-frags/#naming-time) standard.
```
tomrossi7, line 130: From looking at that media-frag document, Is this really a `start` time or is it more like an `interval`? Are you expecting a format like "start_seconds, end_seconds"?

README.md
```diff
@@ -116,6 +116,21 @@ this larger namespace.  But, we don't want to be so general that the spec become
    specified by the RSS \<language\> element.  If the rel="captions" attribute is present, the linked file is considered to be a closed captions file, regardless of what the
    mime type is.  In that scenario, time codes are assumed to be present in the file in some capacity.
 
+- **\<podcast:chapters/>**
+
+   Channel (optional | single)
+
+   This element lists optional `<podcast:chapter/>` item elements.
+
+- **\<podcast:chapter start="[(Normal Play Time)]" title="[(string)]" href="[(uri to content)]" type="[mime type]"/>**
```
MartinMouritzen, line 125: I feel that we could merge the idea of "rich content" and the chapters by having title be not-required. If the title attribute is not present, then this becomes a "sub chapter" which only purpose is to display the content present in the href - in this case we would also need an optional "end" attribute though (and if not present it is assumed that the chapter lasts until the next chapter, or the end of the podcast if this is the last chapter)

daveajones:
Tell me what the point of the channel element:  `<podcast:chapters />` is.  I'm not sure I understand from the brief description.

daveajones:
> 
> 
> Tell me what the point of the channel element: `<podcast:chapters />` is. I'm not sure I understand from the brief description.

Looks like that may be a typo.  Looking at the Simple Chapters Spec it seems that "chapters" is an item level element.

README.md
normand1: Good question. I'm open to using an interval instead. My reasoning for only using is as follows:
- The Simple Chapter Spec only used Start Time so I figured it would be good to default to that unless an interval was needed (maybe it is)
-  From my experience implementing chapters in a client I simply generate an End Time for a chapter by subtracting 1 sec from the next chapter's start time. I think the advantage here is that you don't need to worry as a client developer that the RSS chapter end times are incorrect or inconsistent. If I can generate them myself then I know the chapters intervals are correct. 
- I think intervals could be represented with just start times by including a chapter for `Start Commercial` and a chapter for `End Commercial` if that use case did arise, but admittedly it's not very clean

These aren't arguments against using a time interval instead, just why I went with `Start Time`. I think intervals would be slightly more complicated to support and implement for clients, but would add some functionality some clients might want. So I'm definitely open to that if we agree that would be useful and worth the minor complexity trade-off.

README.md
normand1: I like that idea. I wonder if instead of making `title` optional, which seems to me part of the minimum info required to have a chapter object, if this `<podcast:chapter/>` element was moved outside of a `<podcast:chapters/>` list it might behave differently and instead act as "RichContent" that's applicable to the whole episode instead of a specific time stamp?

normand1:
@daveajones Yep, sorry I'm a bit of an RSS noob. My goal with the `<podcast:chapters/>` tag was just as a container for 
`<podcast:chapter/>` tags. So it would look something like this in RSS:

```
<podcast:chapter>
   <podcast:chapter start="1:00" title="Podcasting 2.0 Website" href="https://podcastindex.org" type="text/html"/>
   <podcast:chapter start="3:56" title="Check"  out the repo!" href="https://github.com/Podcastindex-org" type="text/html"/>
   <podcast:chapter start="10:07" title="Follow us on Twitter" href="https://twitter.com/PodcastindexOrg" type="text/html"/>
</podcast:chapter>
```
I probably have my terms or syntax wrong though

daveajones:
I'm going to go ahead an merge this, since there is a general feeling that this is a good idea.  We can modify it as we go.

README.md
tomrossi7: Perfect! I just went through our chapter data in Buzzsprout and we only store the start time. You may just want to say the `start` is the start time in seconds for the chapter marker. I think that w3.org article may be more confusing.

tomrossi7:
@normand1 sorry I didn't realize you were using `HH:MM:SS`. Would it be simpler to just use seconds?

`<podcast:chapter start="236" title="Check out the repo!" href="https://github.com/Podcastindex-org" type="text/html"/>`

normand1:
@tomrossi7 
> @normand1 sorry I didn't realize you were using HH:MM:SS. Would it be simpler to just use seconds?

Yeah I think seconds would be simpler. I don't know how widely used this [Normal Play Time](https://www.ietf.org/rfc/rfc2326.txt) standard is used and if it's even really a known standard. I thought it would be good to stick with since it's what the original [Simple Chapters Spec](https://podlove.org/simple-chapters/) used, but a lot has changed since 2012. I was trying to stick to standard conventions, but really just "number of seconds" is a pretty standard convention if you ask me, so I'd be perfectly ok with changing this.

tomrossi7:
@normand1 yeah and I realized that the `<itunes:duration>` uses just seconds as well.

PR #29, opened 2020-10-13
replace guest element with persons

MartinMouritzen:
I see it as more flexible, and can work on both a channel and item level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,18 +179,7 @@ this larger namespace.  But, we don't want to be so general that the spec become
    All attributes are required.
 
 
-- **\<podcast:host href="[url of bio/wiki/blog/etc.]" img="[link to image/headshot]">**[name of person]**\</podcast:host>**
-
-   Channel or Item
-
-   (optional | multiple)
-
-   It identifies a host of a podcast episode if in the Item, or an entire podcast if in the Channel.
-
-   All attributes are optional but recommended for disambiguation and good meta-data for apps.
-
-
-**\<podcast:persons/>**
+- **\<podcast:persons/>**
 
    Channel or Item (optional | single)
 
--- a/README.md
+++ b/README.md
@@ -190,15 +190,20 @@ this larger namespace.  But, we don't want to be so general that the spec become
    All attributes are optional but recommended for disambiguation and good meta-data for apps.
 
 
-- **\<podcast:guest href="[url of bio/wiki/blog/etc.]" img="[link to image/headshot]">**[name of person]**\</podcast:guest>**
+**\<podcast:persons/>**
 
-   Item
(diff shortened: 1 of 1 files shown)


MartinMouritzen:
I did consider using the http://microformats.org/wiki/hcard standard as a child element, but I think having name, role, img and href is much simpler. The rest of the hCard standard is addresses, phone numbers etc., which are mostly irrelevant for podcasts.

README.md
```diff
@@ -190,15 +190,20 @@ this larger namespace.  But, we don't want to be so general that the spec become
...
+**\<podcast:persons/>**
 
-   Item
+   Channel or Item (optional | single)
 
-   (optional | multiple)
+   This element lists optional `<podcast:person/>` item elements that reflects persons of interest for the show.
 
-   It identifies a guest in a podcast episode.
+- **\<podcast:person name="[(string)]" role="[(string)]" img="[(uri of content)]" href="[(uri to website/wiki/blog)]"/>**
```
tomrossi7, line 199: Would this eliminate the need for the `<podcast:host>` tag? Instead it would just be a `role="host"`?

MartinMouritzen: Ah, yes. It would solve that as well. I will modify the PR.

tomrossi7:
One thing I didn't pick up on is the introduction of a type of collection tag with `persons`:

```
<podcast:persons>
   <podcast:person>
   <podcast:person>
   <podcast:person>
<podcast:persons>
```

We aren't doing that in other places where multiple tags can exist. @daveajones we may want to adopt a standard approach to whether or not we collect them under a parent.

MartinMouritzen:
> One thing I didn't pick up on is the introduction of a type of collection tag with `persons`:
> 
> ```
> <podcast:persons>
>    <podcast:person>
>    <podcast:person>
>    <podcast:person>
> <podcast:persons>
> ```
> 
> We aren't doing that in other places where multiple tags can exist. @daveajones we may want to adopt a standard approach to whether or not we collect them under a parent.

I am actually leaning towards you having the right approach with transcripts. I feel like there's probably a reason why they didn't go with
<items>
 <item />
 <item />
</items>
in the RSS spec, but just a lot of <item>'s.

So I feel by not having the parent element we are being "more in spirit" with the RSS 2.0 spec.

daveajones:
Yes, I very much don't like grouping things under a parent tag unless it's just obvious to everyone that it makes things easier.

daveajones:
I have no strong feeling about it.  If "person" tag with a "role" attribute makes implementation simpler than host/guest tags I'm all for it.  I'd just say, @MartinMouritzen get rid of that parent tag and just make it "person" everywhere.

PR #33, opened 2020-10-14
<podcast:chapter> Start Time Type Update

normand1:
At @tomrossi7 's recommendation I'm proposing a change from using the lesser known [Normal Play Time](https://www.ietf.org/rfc/rfc2326.txt) type to use seconds since the start of podcast instead represented in integers. This is a simpler way of representing a time within a podcast and will be simpler for clients to adopt.

This also aligns with the type used by the itunes tag: `<itunes:duration>`

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,14 +110,14 @@ this larger namespace.  But, we don't want to be so general that the spec become
    mime type is.  In that scenario, time codes are assumed to be present in the file in some capacity.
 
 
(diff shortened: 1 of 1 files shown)


daveajones:
> 
> 
> Ugh. I just realized we have had some issues with people needing to use a fraction of a second. You may want to let the host provide a number as either int or float. The aggregator could do with it what they want?

Yes, let's make sure the spec aligns with what's happening in the real world already.  Good catch.

PR #34, opened 2020-10-14
"Better" formatting

MartinMouritzen:
Very subjective, but I find it easier to read. I also removed a lot of line breaks that were not fitting the text, as well as adding a few new ones for easier reading.


MartinMouritzen:
Ah, I see that Github is formatting things a bit different than the markdown editor I was using, so it definitely did not help readability. I will give it another pass.

PR #36, opened 2020-10-14
Including the `podcast:locked` tag in the namespace declaration.

tomrossi7:
Buzzsprout is looking at implementing this tag so I propose we formalize its declaration.


daveajones:
Done.  Thanks Tom.

PR #41, opened 2020-10-15
Highlighting apps that have implemented elements

snookfin:


tomrossi7:
I love this idea! By highlighting those apps that are implementing these elements we can motivate others to jump on board! @daveajones we may want to highlight this right from the home page. Maybe a list of apps/hosts/players that have implemented at least one of the proposed tags with a link to this more detailed list?

daveajones:
Excellent idea.  Will do.

Issue #40, opened 2020-10-15
XML Syntax Guidelines


snookfin:
I get that XML is ugly, but this is the language of RSS. Therefore, we should be writing the best XML possible.

**Current syntax:**
`<podcast:funding platform="[service slug]" title="[user provided note (string)]">[url for the show at the platform]</podcast:funding>`

**Proposed syntax:**
`<podcast:funding platform="[service slug]" url="[URL to funding platform]">[user provided content to be linked]</podcast:funding>`

The guiding principle here is that XML is intended for computers and people. The content that's intended for computers, should be included in the element. The content that's intended for people, should be included between the elements opening and closing tags.

Empty elements must still be closed. To be as concise as possible, I propose we agree to the short-hand syntax (I know it's gross).

**Empty element syntax:**
`<podcast:transcript url="https://podcastindex.org/ep0002/transcript.json" type="application/json" language="es" rel="captions" />`

tomrossi7:
> The content that's intended for computers, should be included in the element. The content that's intended for people, should be included between the elements opening and closing tags.

I agree that should be our guiding principle for attributes vs. content.

> Empty elements must still be closed. To be as concise as possible, I propose we agree to the short-hand syntax 

The only thing we really have to agree on here is what convention to follow in our documentation. Whether we agree to it or not, I'm pretty sure its part of XML and all XML parsers should respect the self-closing tag.

cisene:
That would be the exact opposite to "Goal #2 - Minimize Attributes and Sub-elements" ..

snookfin:
@cisene - Not really. If you look at my example, you'll see I'm not adding additional attributes. I'm just moving things around. The content to be linked should live in the tag and the URL should be the attribute.

daveajones:
@cisene Considering booting goal 2 from the doc.  It's a good aim.  But that ship has sailed completely now.  We're attribute city.  :-)  I'm ok with it.  This is the industry designing the spec they need.

@snookfin Shoot me a PR and I'll merge.  Also, we probably need to review and make sure there is consistency across the namespace with url="" and href="" so we can pick just one of those to stick with.

tomrossi7:
@daveajones Goal #2 may just need to be tweaked. We should be hesitant to add code *anywhere* in the RSS feed. Whether its attributes or just introducing tags that may not be useful. Something more along those lines maybe?

cisene:
[...]

PR #42, opened 2020-10-15
Syntax updates as requested

snookfin:
As discussed here: https://github.com/Podcastindex-org/podcast-namespace/issues/40#issuecomment-709557157

url="" is used everywhere except for image which use href="". This is consistent with existing standards (iTunes namespace).


daveajones:
Perfect.

PR #52, opened 2020-10-17
Update README.md

jamescridland:
Removing combative language. We don't need to shame anyone into making changes - we need to enthuse and excite them...


daveajones:
Good catch.

PR #102, opened 2020-11-08
Fix markdown close on podcast:indexers note + millisat-bitcoin change

AbelLykens:


AbelLykens:
Should have worked with branches, now the bitcoin-millisat commit is also included here, sorry.

PR #119, opened 2020-11-18
Update contributors.txt

Besiuk:
You guys asked to add ourselves, even if it's only a little contribution on podcastindex.social so here I am :)
keep up the good work!


daveajones:
Yes. Thanks!

PR #127, opened 2020-11-23
Update Element Support

tylerlacy:
PodServe.fm has added support for the locked tag. Not only in our feeds, but block imports if feeds are locked. The funding tag is also enabled at PodServe.fm


daveajones:
Excellent. I’ll update the apps page to reflect on podcastindex.org.

daveajones:
Don’t forget to add yourself to the contributors list as well. Support == contribution.

PR #139, opened 2020-12-07
Update README.md

jamescridland:
Suggestion to remove the location tag from here for now, to save confusion.


daveajones:
This is a good idea.

PR #143, opened 2020-12-11
Change namespace URI

### LINKED ISSUE #113: xmlns attribute
I don't know if this would be a breaking change for consumers, but I think we should change the xmlns (a.k.a THE namespace) attribute.

It currently points to a GitHub domain (https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/1.0.md), I have the feeling it would be better to use a URI that uses the podcastindex.org domain.

I suggest something like https://podcastindex.org/namespace/1.0, and setup a redirect of this url to the documentation in GitHub.

This way you have more flexibility of what you want to use the URL for (like if you move the documentation elsewhere).

### Refs

https://www.w3schools.com/XML/xml_namespaces.asp

> When using prefixes in XML, a namespace for the prefix must be defined.
> 
> The namespace can be defined by an xmlns attribute in the start tag of an element.
> 
> The namespace declaration has the following syntax. xmlns:prefix="URI".

> Note: The namespace URI is not used by the parser to look up information.
> 
> The purpose of using an URI is to give the namespace a unique name.
> 
> However, **companies often use the namespace as a pointer to a web page containing namespace information.**

> A Uniform Resource Identifier (URI) is a string of characters which identifies an Internet Resource.
> 
> The most common URI is the Uniform Resource Locator (URL) which identifies an Internet domain address. Another, not so common type of URI is the Uniform Resource Name (URN).

coldacid:
This changes the XML namespace to `https://podcastindex.org/namespace/1.0` while retaining backwards compatibility to the old namespace. It also provides guidance regarding the XML prefix for the namespace's tags while reminding implementers that they should respect alternate prefixes for the same namespace URI.

Resolves #113.

### DIFF
--- a/docs/1.0.md
+++ b/docs/1.0.md
@@ -2,8 +2,12 @@
 
 A wholistic RSS namespace for podcasting that is meant to synthesize the fragmented world of podcast namespaces. As elements are canonized, they will be added to this document so developers can begin implementation. The specifications below are considered locked and the team will prioritize backward compatibility. We are operating under the [Rules for Standards-Makers](http://scripting.com/2017/05/09/rulesForStandardsmakers.html).
 
-If your application generates RSS feeds and you implement one or more elements below, you will need to link this DTD in your XML.
-`<rss version="2.0" xmlns:podcast="https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/1.0.md">`
+The namespace for this extension is `http://podcastindex.net/namespace/1.0`. Clients which recognize this namespace must also recognize `https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/1.0.md` as identical. The suggested tag prefix for use in XML is `podcast`, but clients should support alternate prefixes for this namespace. If your application generates RSS feeds and you implement one or more elements below, you will need to link this DTD in your XML:
+
+```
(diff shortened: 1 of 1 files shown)


docs/1.0.md
```diff
@@ -2,8 +2,12 @@
...
 
-If your application generates RSS feeds and you implement one or more elements below, you will need to link this DTD in your XML.
-`<rss version="2.0" xmlns:podcast="https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/1.0.md">`
+The namespace for this extension is `https://podcastindex.org/namespace/1.0`. Clients which recognize this namespace must also recognize `https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/1.0.md` as identical. The suggested tag prefix for use in XML is `podcast`, but clients should support alternate prefixes for this namespace. If your application generates RSS feeds and you implement one or more elements below, you will need to link this DTD in your XML:
+
+```
+<rss version="2.0" xmlns:podcast="http://podcastindex.net/namespace/1.0">
```
saerdnaer, line 8: `http` or `https`?

daveajones: HTTPS for sure.

PR #197, opened 2021-02-25
Add Anytime to funding and chapters

amugofjava:
Hi @daveajones,

This adds Anytime to Funding and Chapters.

Thanks.


daveajones:
Added manually because of conflicts.

PR #206, opened 2021-03-08
Hive Value Block added plus minor notes on Lightning block

brianoflondon:
I've added a version of the Value block for Hive which follows the work I'm doing. 

I've fixed a couple of things in the Lightning block:
1. no mention made of the human readable `name` attribute.
2. suggested is still an issue as the spec says integer milisatoshis but everyone is still using a float denominated in full BTC.

### DIFF
--- a/value/value.md
+++ b/value/value.md
@@ -262,11 +262,12 @@ For the `<podcast:value>` tag, the following attributes MUST be used:
 
  - `type` (required): "lightning"
  - `method` (required): "keysend"
- - `suggested` (optional): An integer representing millisatoshis.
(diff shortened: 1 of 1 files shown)


brianoflondon:
I've just re-looked at the entire document and I realise now the structure and my comment adding name to the lightning block section is probably wrong as I wasn't looking further up the page to the non-option values.

The confusion over milisatoshis still remains.

brianoflondon:
@daveajones I think this is good to go, I only added one thing to lightning spec about minutes.

I still think the spec is confusing over saying suggested is in milisatoshis but then having a decimal number in the example that looks like its denominated in full Bitcoin.

Issue #116, opened 2020-11-16
<podcast:value> Per-item override for value tag


AbelLykens:
Thoughts:

 - Allow for overwriting the "main" (root) value tag per episode
 - If I get a guest in my podcast I might want to share revenue for that episode with the guest
 - If there is no value tag in the item, use the mail value tag

daveajones:
Excellent idea. I saw an example of how this would be useful over the weekend.

johnchidgey:
I'd like to add a vote for this too please. On Pragmatic I can foresee wishing to value-split each episode to guests that come on the show. Since there's different guests each episode then this would be different for each episode. Hence I would suggest that the primary Value tag split that would ordinarily go to the host, would then be split between the host and the nominated guest(s) on an episode by episode basis where nominated. Need to ensure the top-level splits are honoured first (podcast app, hosting services etc) before sub-splitting at the Item (episode) level. Just my 2c.

satoshisstream:
#126 partly helps. We could do an overall value tag as we have now. 

Then if there is a value block in an episode, use that to override general tag.

If there is a value block in a CHAPTER, use that to override all.

Most specific use is leading...?

daveajones:
Let’s work this into phase 3. It should be fairly straightforward and we’re already making a change of chapter format to add location data. Seems like a good time to do it.

vv01f:
this maybe just prose what how I understand this, trying to grasp how this will take effect …

1. value data in the channel (maybe that you mean with root value tag) should be taken over defaults in the app (e.g. `suggested` attribute)
2. customized value data by the user in the app should override app defaults as well as any data on the feed
3. per item value data should override the channel (could be mistanken for solving this issue already)
4. per chapter valueOverride #216 should dominate even the item value data
  + this feels like a workaround for my use case: send custom metadata for this time
  + this is also enabling to have a short part (snippets for a mix) being recognized

what I miss:
* a) which data is to be overridden and how should it look like?
* b) I guess it might not be sufficient to manipulate the data given by the value block itself yet.
* c) this might be not just a single attribute but a few instead.
* d) the attributes `customKey` and `customValue` could already solve for a single keyvalue of metadata if that was recognized, just I lack to find an example or statement on the spec for this
* e) what keyvalue pairs need to be addressed with this or which definitely should not?

my usecase in mind was:
[...]

PR #216, opened 2021-03-28
Suggested valueOverride

satoshisstream:
Allow for an overriding valueblock for specific chapters, for example when music is played.

### DIFF
--- a/chapters/exampleComplex.json
+++ b/chapters/exampleComplex.json
@@ -22,7 +22,7 @@
       "startTime": 410,
       "title": "Namespace",
       "img": "https://example.com/images/namepsace_example.jpg",
-      "url": "https://github.com/Podcastindex-org/podcast-namespace"
+      "url": "https://github.com/Podcastindex-org/podcast-namespace",
     },
     {
       "startTime": 3990,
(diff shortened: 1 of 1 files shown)


AbelLykens:
This should help with  #116

chapters/exampleComplex.json
```diff
@@ -22,14 +22,21 @@
...
       "title": "Just Break Up",
       "img": "https://example.com/images/justbreakuppod.png",
       "url": "https://twitter.com/justbreakuppod"
     },
+     {
+      "startTime": 4200,
+      "title": "Played song by artist",
+      "img": "https://i.discogs.com/R-249504-1334592212.jpeg?bucket=discogs-images&fit=contain&format=auto&height=600&quality=90&width=600&signature=wMI9I0mHkbVxkkryQrN1JkkzhwFsrereuom9Lmfa92w%3D",
+      "url": "https://www.discogs.com/Rick-Astley-Never-Gonna-Give-You-Up/master/96559"
+      "valueOverride": "<podcast:value type=\"lightning\" method=\"keysend\" suggested=\"0.00000005000\"><podcast:valueRecipient name=\"Performing Artist\" type=\"node\" address=\"03c457fafbc8b91b462ef0b8f61d4fd96577a4b58c18b50e59621fd0f41a8ae1a4\" split=\"100\"/></podcast:value>"
```
dergigi, line 38: Any reason to include the explicit "Override" in the name? 

I think simply calling it `value` would be clear and in line with the `podcast:value` block definition, which states:

> The <podcast:value> tag designates the cryptocurrency or payment layer that will be used, the transport method for transacting the payments, and a suggested amount denominated in the given cryptocurrency.
> This element can exist at either the <channel> or <item> level.

daveajones: I agree with this.

PR #220, opened 2021-03-31
Add WebVTT example/documentation to transcripts

### LINKED ISSUE #217: WebVTT support for podcast:transcript?
Working with some web technologies, I've come across the WebVTT subtitles/caption format.  It's very similar to SRT and seems to be increasing in use on the web.

https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API

agates:
Closes #217


daveajones:
Perfect. Thanks Alecks.

PR #221, opened 2021-04-02
Change "application/srt" to "text/srt" to be consistent with Examples.

johnchidgey:
Looking through here: https://www.iana.org/assignments/media-types/media-types.xhtml#text there’s no specific type for SRT, but there is for text/vtt and text/vnd.dvb.subtitle so I think we should ammend the PC2.0 docs: https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/1.0.md Currently says: "type: Mime type of the file such as text/plain, text/html, application/srt, application/json” but application/srt should be text/srt I think.

### DIFF
--- a/docs/1.0.md
+++ b/docs/1.0.md
@@ -29,7 +29,7 @@ Multiple
 #### Attributes
  - **url:** URL of the podcast transcript.
 
- - **type:** Mime type of the file such as `text/plain`, `text/html`, `application/srt`, `application/json`
+ - **type:** Mime type of the file such as `text/plain`, `text/html`, `text/srt`, `application/json`
 
  - **language (optional):** The language of the linked transcript. If there is no language attribute given, the linked file is assumed to be the same language that is specified by the RSS `<language>` element.
 


saerdnaer:
We use `application/x-subrip` and `text/vtt`... but yes `text/srt` would be nice, but has to introduced by IANA first.

joksas:
It would be nice to have official IANA support, but I think @johnchidgey is also referring to the fact that the specification page itself has inconsistencies.

The folowing line uses `application/srt`
https://github.com/Podcastindex-org/podcast-namespace/blob/c311bcad76e5779bb6474dfa7291fc3d87839510/docs/1.0.md#L34

While the following line uses `text/srt`
https://github.com/Podcastindex-org/podcast-namespace/blob/c311bcad76e5779bb6474dfa7291fc3d87839510/docs/1.0.md#L46

tomrossi7:
We went back and forth on this originally and ended up `application/srt`.  Since its already in production for folks, I would prefer to standardize on that unless someone has a compelling reason for `text/` rather than `application/`

joksas:
I think [this](https://github.com/gabriel-vasile/mimetype/pull/232#pullrequestreview-853753882) is a strong argument *against* using either `application/srt` or `text/srt`. It's going to be difficult to convince developers maintaining mimetype detection tools to support mimetypes that go against the standards of IANA. Of course, internally, podcast apps could change the mimetype to whatever they wish, but that's kind of convoluted, no?

Issue #228, opened 2021-04-09
integer | float for value suggested 


vv01f:
Currently the data type for the suggested amount is inconsistent. for XSD both would fit in decimal.

* int is mandatory for lightning, but there is a float in the example
* float is mentioned in general and seems to be in use

I guess the integer lost due to common practice?

daveajones:
Agreed.  Will fix.

PR #230, opened 2021-04-14
Merge branch 'main' of https://github.com/brianoflondon/podcast-namespace into main

brianoflondon:
Chaned the Suggested value from integer to float and added example of 5 Sats (which is the default value in the No Agenda feed).


dergigi:
I think the following paragraph now needs updating:

> The "suggested" amount is just that. It's a suggestion, and must be changeable by the user to another value, or to zero. The suggested amount should always be given in the smallest denomination available within the payment protocol being used. For instance, with Lightning it is given in millisatoshis.

See also issue #228

PR #242, opened 2021-04-25
Add Podlove Publisher & Podcat to support list

saerdnaer:


saerdnaer:
On second thought: Maybe we should simply delete the file and change all references to the new solution on https://podcastindex.org/apps (or similar)

PR #271, opened 2021-07-22
Fix it's vs. its in docs

misener:
Replaced several instances of "it's" with "its" in docs (https://www.grammarly.com/blog/its-vs-its/)


daveajones:
Thank you.  It's a perpetual problem with my writing.

PR #272, opened 2021-07-23
Value: Add `<Item>`-Level Override Example

dergigi:
I have updated the existing example slightly to make clear that the `<podcast:value>` block is set on a `<channel>` level.

Further, I have introduced a second example, which shows that a different `<podcast:value>` block can be set for each `<item>`, i.e. for each podcast episode.

I hope the example as well as the use-case is clear. It is again an example from the real world, showcasing a real use-case that we want to implement, namely including every guest in the payment stream. 

Please let me know if anything is unclear or the wording (or the examples) could be improved.

### DIFF
--- a/value/value.md
+++ b/value/value.md
@@ -341,6 +341,13 @@ Since the value block is defined at the `<channel>` level, it applies to every p
 
 ##### Example: `<Item>` Override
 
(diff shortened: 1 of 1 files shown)


daveajones:
This looks great Gigi. Very clear.

PR #274, opened 2021-07-29
Update value example and description to include AMP

dergigi:
To quote from the [lnd 0.13-beta announcement](https://lightning.engineering/posts/2021-06-17-lnd-v0.13/):

> AMP is the successor to the widely used “keysend” protocol as it enables payments to be sent in multiple shards similar to Multi-Path Payments (MPP). Unlike MPP, the receiver cannot pull the payment until all the payment shards arrive. This trait makes the new payment type more useful for payment-splitting aware probing.

Nothing in the specification has to change to support AMP. Just specify `amp` as the `method` instead of `keysend`.

To quote from the announcement again:

> Sending a spontaneous AMP payment requires specifying the amount and pubkey of the destination

I have linked to [Multipath Payments](https://bitcoinops.org/en/topics/multipath-payments/) page of Bitcoin Optech. It includes all relevant information.

---

Additional resources on AMP that might be useful:

* https://wiki.ion.radar.tech/tech/research/atomic-multi-path-payments
* https://bitcoin.stackexchange.com/a/89476

### DIFF
--- a/value/value.md
+++ b/value/value.md
@@ -265,12 +265,12 @@ API that exposes a compatible process.
 
 Currently, development is centered fully on [Lightning](https://github.com/lightningnetwork) using the "keysend" protocol.  Keysend allows for push
 based payments without the recipient needing to generate an invoice to receive them.
+Another method to send spontaneous payments via Lightning is AMP, atomic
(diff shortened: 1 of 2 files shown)


daveajones:
Looks great.

PR #275, opened 2021-07-29
Update element-support.md

montekaka:
JustCast.com has added support for the chapters element.


daveajones:
Beautiful.  You'll probably want to update:  https://github.com/Podcastindex-org/web-ui/blob/master/server/data/apps.json as well.  That's what the public /apps page pulls from on Podcastindex.org

PR #283, opened 2021-08-17
Podping tag edit and new tag for Hive account name

brianoflondon:
Right now with 3speak I am deducing the Hive account name from the name of the podcast but in future I'd like to be able to have a dedicated tag in the Podcast namespace so a podcast anywhere can identify the Hive owner and that would allow routing of value for value payments without reference to the "name" tag in a value block.

The Value block would still have to route to a service that provides onward transfer of Lightning to Hive (at present that's being run by Brianoflondon only) but the goal is to open source this bridge code and allow anyone to run this servce so it can be distributed.


satoshisstream:
I suggest making an optional generic GUID-like identifier, not Hive-bound? 

If I am understanding you correctly, I suggest using `NAME` in `7629169`, which should be filled with the name from the split. If apps would use this, you could link the _unique_ name from the split to a receiver?

brianoflondon:
I am using Name right now (and a fall back of looking in the Podcast Name) but it means that sometimes these fields (which are surfaced in app UI), won't be able to be something the podcaster wants. I'd rather, for example, the Name field and much of the V4V app UI said "Brian of London" but behind the scenes that would route that to `@brianoflondon` on Hive without having to have `brianoflondon` in either the Podcast Name or the Value block Name field.

I'm not such a great example, because I've got a great Hive name, but not everyone has a perfect alignment. 

The field here only comes into use if/when the Podcast Index namespace proposal for a dedicated Hive Account Name tag is accepted.

satoshisstream:
I suppose on your receiving end (Hive), you register what podcast is whose user, so you have a unique key [feedid-name], right?

The thing you propose does sound useful if you don't want to do administration, and just want a unique receiver ID.

We could turn this into a receiver identifier (`unique-identifier`), which can be used in multiple podcasts/value blocks, and any payment to that GUID is linked to a unique user (wherever the value is sent to)?

Hive users would set the `unique-identifier` to their Hive name. The receiver would understand this, because value is sent to the key provided by the user in the value block.

Satoshis.Stream users would set the `unique-identifier` to an identifier given by us. We would understand the `unique-identifier` because we have given this identifier to the user.

This would result in a reusable key, less registration.

brianoflondon:
I'm all the time thinking what would be decentralised. Having a centralised (even if I store it on Hive) list of this GUID or this URL links to this Hive account is bad. Outside the world of crypto and geeks, asking people to paste `unique-identifiers` around when they already have something like a Hive account just makes for mistakes and complexity.

Having the Hive account name (just like having a Lightning Keysend address) in the feed is good and decentralised. 

The other place to put this is as an attribute only in the value block (alongside name), but there's lots more we can do when a podcast links itself directly to Hive at the top level which is why I'm making this proposal now.

AbelLykens:
`unique-identifier` would be decentralized, and provided by the receiving service. It's the receiving service which would link this to a user.

The receiving service could even provide multiple `unique-identifier`s per user, or have the user provide their own `unique-identifier`s, as long as that receiving service remembers what `unique-identifier`-received value to send to whom...?

satoshisstream:
`unique-identifier` would be decentralized, and provided by the receiving service. It's the receiving service which would link this to a user.

The receiving service could even provide multiple `unique-identifier`s per user, or have the user provide their own `unique-identifier`s, as long as that receiving service remembers what `unique-identifier`-received value to send to whom...?

Issue #300, opened 2021-10-07
Allow multiple `podcast:value` tags per podcast/item


dellagustin:
Moving forward, it is possible that a podcast is able to receive _value-for-value_ support using other means then Bitcoin over lightning. 

A simple example would be Litecoin over lightning.
I think this could be funcional already if there is tooling to support it, as existing nodes (lnd at least) are already capable of operating with Litecoin.

One consideration is priority and fallback strategy  i.e. which one should be used if both sender and receiver support more then one method.

brianoflondon:
> The spec [here](https://github.com/Podcastindex-org/podcast-namespace/blob/24e447a16acfe9c73cf4040adb443f64dbeb493d/value/value.md#value-element) states:
> 
> > Currently, there can be only a single copy of this element at each level.
> 

Interesting, I remember discussing multiple value blocks and I had a Hive block in my feed for a while but had forgotten it got set to allow only one.

I would be in favour or multiple but I'm also cognizant that at this stage with such thin adoption, it might not help to spread us even thinner.

dellagustin:
Hello @brianoflondon ,

> I would be in favour or multiple but I'm also cognizant that at this stage with such thin adoption, it might not help to spread us even thinner.

I don't think that making it official in the standard will spread us thinner. Most people will, in my opinion, follow with Lightning if that is what is easier to adopt (and in the current ecosystem, that is the case, to my best knowledge).

The main benefit I see in making it official in the specification that multiple `podcast:value` rtags are allowed, despite not having other type then lightning at the moment, is that client developers will already design their clients in a future proof way.

That is what I did when designing podStation, I already consider `podcast:value` as a zero-to-many tag in my internal data model, and I filter for lightning when using it.

Marzal:
This was implemented on https://github.com/Podcastindex-org/podcast-namespace/commit/e22a19566026968e1e88688d7ab9baa1a1e0c1e9

PR #303, opened 2021-10-11
Integrate Nano cryptocurrency in V4V spec, adding documentation

koczadly:
This adds a Value 4 Value specification and associated constants for the [Nano](https://nano.org) cryptocurrency.

If there's another procedure for registering and declaring a cryptocurrency in the V4V spec, please let me know and I'll be sure to do it.

### DIFF
--- a/value/value.md
+++ b/value/value.md
@@ -423,3 +423,37 @@ and guest.
   ...
 </channel>
 ```
+
+#### [Nano](https://nano.org)
+
+For the `<podcast:value>` tag, the following attributes MUST be used:
+ - `type` (*required*): "nano"
+ - `method` (*required*): "sendblock"
+ - `suggested` (*optional*): A float representing the amount in nano (eg. 0.025)
+
+For the `<podcast:valueRecipient>` tag, the following attributes MUST be used:
+ - `type`: "account"
+ - `address`: \<*destination nano_xxx address*\>
+ - `split`: \<*the number of shares*\>
+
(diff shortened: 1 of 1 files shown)


brianoflondon:
There is one constraint and perhaps we do need to lift it: the current Value Block spec says only one type of value block per feed. I always thought it should be more but I backed down on that.

I'm happy to re-add my Hive suggestion too but I'm not in a position to drive adoption of it.

dellagustin:
> There is one constraint and perhaps we do need to lift it: the current Value Block spec says only one type of value block per feed. I always thought it should be more but I backed down on that.

Related to that - https://github.com/Podcastindex-org/podcast-namespace/issues/300

daveajones:
@zhyatt Could you resolve the conflicts or do a new PR on this one?  I got behind in merging some other ones and it looks like it messed you up.

koczadly:
@daveajones I have resolved the conflicts.

The appendix section appears to be very lightning specific, and is a nested heading of the lightning section. In my commit I have kept the appendix together under the lightning heading, but this may be out of place?

I do think the document could perhaps do with some restructuring. For instance, the value override example is shown under the Supported currencies > Lightning heading, whereas it's more of a general example.

It may also be worth hiding the supported currencies section within `<details>` blocks (if they're still supported?), allowing the reader to expand only the relevant sections.

daveajones:
Thanks!!

PR #307, opened 2021-10-16
Updated podcast:value specification for custom records

dellagustin:
This commit makes the usage of `customKey` and `customValue` more specific.

The change to the spec was made based on the known users for these attributes (LNPay and HIVE gateway), but it is not necessarily backwards compatible.

## Why is this change necessary?

Client applications that use the information on the `podcast:value` tag must know what to with its content.

Imagine one receiver sets `customValue` to raw binary data encoded in `hex`, and expects it to be sent in the payment as raw binary data.

Imagine another receiver sets `customValue` to raw binary data encoded in base64, and expects it to be sent in the payment as raw binary data.

Imagine now that other receiver sets `customValue` to a string, and expects it to be sent in the payment as a string encoded in UTF-8 (which currently known users do).

There is no way a client could know what do to with `customValue` (how to decode it an re-encode it), without specific knowledge about how each receiver uses this field.

Although this is possible, it is counter productive for client development, as clients should have to know as little as possible
about different application specific protocols used by the receivers.

## References

- [BOLT #1: Base Protocol](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md)
- [TLV record registry](https://github.com/satoshisstream/satoshis.stream/blob/67f5db14fac8260341f0a4733733e3b20b75faa7/TLV_registry.md#tlv-record-registry) by @satoshisstream (permanent link to current version)
- Issue in podStation - https://github.com/podStation/podStation/issues/283
    - analysis and backtracking of the usage of `customValue `in the comments

### DIFF
--- a/value/value.md
+++ b/value/value.md
@@ -290,15 +290,21 @@ For the `<podcast:valueRecipient>` tag, the following attributes MUST be used:
  - `address`: \<the destination node's pubkey\>
  - `split`: \<the number of shares\>
 
-If the receiving Lightning node, or service, requires a custom record or meta-data of some sort to be passed along with the payment
-the `customKey` and `customValue` can be utilized as follows:
+If the receiving Lightning node, or service, requires application specific data to be sent with the payment in the lightning  message `extension` (a _TLV stream_, see the Appendix section), the `customKey` and `customValue` can be utilized as follows:
 
  - `type`: "node"
  - `method`: "keysend" or "amp"
- - `customKey`: \<key name\>
- - `customValue`: \<value\>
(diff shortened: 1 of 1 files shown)


dellagustin:
@brianoflondon , @tkijewski , which encoding do you expect when decoding the value from the custom records?
utf-8 works, as, to my best knowledge you use characters within the ASCII range - within this range UTF-8 is compatible with other encodings such as Windows-1252 and ISO 8859–1, so in general it should not be a problem, but in case you have anything against using UTF-8 in the spec, please comment.

PR #314, opened 2021-11-19
Rename fields, add Networks

benjaminbellamy:


daveajones:
Hey @benjaminbellamy I would suggest incrementing the version number and date on the document as well.

PR #319, opened 2021-12-19
Clarify podcast:images srcset url requirements

johnspurlock:
Specify that absolute urls for the `srcset` attribute are required for `<podcast:images>`, even though the HTML5 `srcset` attribute allows relative urls.


daveajones:
Nice catch. Thanks.

Issue #239, opened 2021-04-23
Proposal: universal method for premium subscriptions


theDanielJLewis:
After Apple's announcement about premium subscriptions (which many had expected), I've been thinking of a way that we could build a more universal "subscription" method that app developers could support.

## Goals

1. To let podcasters monetize their podcasts directly via subscriptions, donations, exclusive access, bonuses, and more.
2. To provide podcast-app developers a framework to launch their own support and earn some extra profit, too.
3. to create a universal standard that would be easy for podcasters and podcast-consumers, and support the freedom of choice for consumption platforms.

## Proposal: `<podcast:exclusive>`

My proposal is for a new `<podcast:exclusive>` tag and framework. The could be used at the channel level to lock an entire feed, at the item level to lock individual episodes, or even within an item to lock specific portions of content or additional bonuses.

While I think some of my idea might be invalid RSS or otherwise rough, I'll share it anyway to help inspire ideas for the problems I'm trying to solve.

### Code concepts

There would be two ways to apply the "exclusive" feature: tag and attribute.

Tag: `<podcast:exclusive roles="bronze,silver,gold">`

Attribute: `<enclosure exclusive-url="…" … />` or `<title exclusive="90-minute episode">60-minute episode</title>`

The idea is that the exclusive tag would wrap or mark whole items as exclusive, while the attribute would mark smaller pieces as exclusive.

A probably-invalid method for making an exclusive episode in a public feed could be like this:

```xml
<exclusive role="subscriber">
  <item>
    …
  </item>
</exclusive>
<item>
  …
</item>
```

My theory with this example is that it hides the item from unauthenticated users and from default RSS-reading apps (podcast apps or feed readers). This is _not_ creating any security, only acting as a way to prevent exclusive content from showing in normal feed-reading places.

Or if you want the episode to be visible, only not downloadable, you could add the `<podcast:exclusive>` tag inside the item.

And if the only difference for the exclusive item is the media (like maybe an ad-free version), you would use the exclusive attribute on the enclosure (or maybe alternate enclosure).

### Security and Authentication

Media would need to be stored in a secured method, such as with the Amazon S3 protocol. So the URL in the feed would be a normal URL that works for everyone, but the podcast app would have the access key stored to authenticate and access the media.

This _would_ require some cooperation with podcast-publishing or hosting tools, to host with the authentication requirement. But neither the podcaster nor the host would _have_ to manage the memberships, roles, and payments. That could be left to app developers.

[...]

PR #323, opened 2022-01-12
Proposal for a bonusItem element

francosolerio:
To encourage adoption of Value4Value I put together a proposal for the addition of a <bonusItem> element for the namespace.

It gives the possibility to publish bonus content available only to listeners who reach a minimum monetary contribution. 

This would encourage listeners contributions, the adoption of apps that support Podcasting 2.0, and the implementation of the spec by other apps. Non-P2.0 apps wouldn't display bonus content, resulting in a competitive advantage for those who do.

### DIFF
--- a/proposal-docs/bonusItem/bonusItem.md
+++ b/proposal-docs/bonusItem/bonusItem.md
@@ -1,0 +1,53 @@
+# The "podcast:bonusItem" Specification
+
+<small>Version 1.0 by [Franco Solerio](https://github.com/francosolerio)</small><br>
+<small>January 10th, 2022</small>
+
+## Purpose
+
+With Value4Value podcasters can give their audience the possibility to send payments in response to the action of listening to a show. After a few months from its invention many shows already demonstrated that audiences embrace this possibility and reward podcasters, creating a virtuous loop that favours the production of more and better content. It is also evident that single listeners contribute in very different degrees, from nothing to big amounts. This can be attributed to many reasons, just to name a few: use of apps that still don't support Podcasting 2.0, the listener's financial condition, education ad good will, the podcaster's ability to stimulate contribution both by improving the quality of their show and istigating the listeners to contribute.
+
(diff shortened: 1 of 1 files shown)


theDanielJLewis:
This sounds similar to the ideas I put forth in #239, which would allow any content to be marked as exclusive, granting access only under certain conditions (like V4V, website membership, OAuth, etc.).

francosolerio:
Thank you @theDanielJLewis, I didn't remember about that proposal. What I was thinking here is a softer approach. No password, no subscription, no private/public keys or DRM. It's not meant to replicate the subscription experience Apple or Spotify propose, but just to make it possible for podcasters to thank contributors rewarding them with some bonus content.

Obviously anybody could code an app, or a simple script, that downloads bonus items without contributing. This is all based on the same "gentlemen agreement" philosophy that is implied by Value4Value: you're not forced to contribute, but we kindly ask you too. And as a bonus if you do we reward you with some bonus content.

jamescridland:
I have to say, I love this approach - it relies on podcast apps supporting Podcasting 2.0 and v4v, and allows some validation of that approach.

We shouldn't limit bonus content to being just audio. I'd quite like the idea of a badge or something, which could appear in a user's profile - "I'm a Podland contributor", for example. I think we can deliver that, though would quite like to consider a badge as a unique item type, with some standards around sizing and delivery.

daveajones:
> I'd quite like the idea of a badge or something, which could appear in a user's profile - "I'm a Podland contributor", for example.

I love that. Could even be an NFT. Not really something I’m into, but lots of people are.

adamc199:
Excellent idea.

theDanielJLewis:
I like the approach. I would still call this "exclusive," even though it's not behind a subscription. And some or all exclusives could be unlocked by V4V, subscription, donation, or anything else.

When I mention keys, OAuth, and such, that would all have to be completely invisible to the podcaster; just give and get.

PR #331, opened 2022-01-22
use proper mime type for srt files #221

saerdnaer:
consistently use `application/x-subrip` for srt, unify order


daveajones:
Thanks!!

PR #373, opened 2022-06-08
Add RSS Blue

joksas:
Hi,

I'd appreciate it if [RSS Blue](https://rssblue.com) could be added. It currently supports the following elements:

- Transcript (uploading a file)
- Locked
- Funding
- Chapters (uploading a file)
- Location
- Value (*one* Lightning node at this point; still need to figure out intuitive UI for adding multiple addresses with different splits)

Example feed that includes elements "Locked", "Location" and "Value" can be found [here](https://rssblue.com/@future-of-computing/feed). Please let me know if there are any issues.


daveajones:
Looks great!

PR #383, opened 2022-08-20
remove nano

callebtc:
This PR removes the cryptocurrency Nano from the spec.

### DIFF
--- a/value/value.md
+++ b/value/value.md
@@ -450,39 +450,3 @@ One part of the message is the `extension`, a TLV (Type-Length-Value) stream. Ap
 
 A community maintained registry of known custom record types and formats, governed by satoshis.stream, can be found at the document [TLV record registry](https://github.com/satoshisstream/satoshis.stream/blob/main/TLV_registry.md). In special, the section _Fields used in customKey / customValue Pairs_ documents the known use cases for the `customKey` and `customValue` attributes.
 
-
-#### Nano
-
-Specification for Value 4 Value integration with the [Nano](https://nano.org) cryptocurrency.
-
-For the `<podcast:value>` tag, the following attributes are applicable:
- - `type` (*required*): "nano"
- - `method` (*required*): "sendblock"
- - `suggested` (*optional*): A float representing the amount in nano (eg. 0.00025)
-
-For the `<podcast:valueRecipient>` tag, the following attributes MUST be used:
- - `type`: "account"
- - `address`: \<*destination nano_xxx address*\>
- - `split`: \<*the number of shares*\>
-
-##### Example
-***Note:** The following is a made up example, and the accounts shown are for demonstration purposes only.*
-
-```xml
-<channel>
-    <podcast:value type="nano" method="sendblock" suggested="0.0001">
-        <podcast:valueRecipient
-            name="John Doe (Host)"
-            type="account"
-            address="nano_3fegpdys1a3ncscmhagn3yxdrf5s9kb8drorsi75399omnwjbcp87cyibzyx"
-            split="50"
-        />
-        <podcast:valueRecipient
-            name="Jane Doe (Guest)"
-            type="account"
-            address="nano_1ero6egquwf3bkh4h5ksf7ghumehn3c131opqe65bsi5q17huwjsnhcp8kxd"
-            split="50"
-        />
-    </podcast:value>
-</channel>
-```


benthecarman (approved):

brianoflondon:
This is an unimplemented example. Why would you remove it? It demonstrates that this is an extensible protocol, not a fixed closed system.

daveajones:
I agree with @brianoflondon. Why remove this?  I don’t see the point.

dergigi:
I think the point is that there are over 9000 "cryptocurrencies" (9651, as of this writing) listed on coinmarketcap, all of which could be added to the spec, in theory.

If you want to be inclusive maybe the way forward would be have a generic altcoin spec, and a separate section that lists altcoin implementations with examples?

brianoflondon:
I originally put Hive in there and removed it because I recognised that Lightning had the mind share and that there are good reasons why Hive is less suitable for having potentially millions of low value disposable wallets (Lighting is good for this, most Layer 1 blockchains won't be). 

Nano actually could be shoe horned into this but nobody will and it serves as a placeholder.

For example. BTC wouldn't work (too slow, too expensive) and neither would ETH and any ERC20 derived tokens on the ETH main chain which make up most of the 9000 on CMC. Could someone propose something on Polygon or any of the other clones of ETH? Yes, but from this point on @daveajones and I would probably say come back with a proof of concept and we'll consider adding it to the spec.

dipunm:
Showing nano as an example serves as endorsement.

Either use a explicitly vague example eg "[crypto address/details here]", or at least stick to the original well known and well respected bitcoin.

Using bitcoin as an example is like using 0 in your unit tests for a number field. No one asks "why zero?" They accept it as a default number unimportant to the test itself.

daveajones:
> ...maybe the way forward would be have a generic altcoin spec, and a separate section that lists altcoin implementations with examples?

In that regard, maybe just leave the nano example there as an example of how it's done.  It could fill the role of generic cryptocoin since I've honestly never even heard of nano.  I'm fine with spawning a new document called value-extensions or some such that just holds all of these if anyone wants to do them (they won't).  That way we don't have to modify the original whitepaper.

jamescridland:
Chiming in: a general complaint of mine about the podcast namespace is that there are many different things in the specification, and many aren't supported. This leads to patchy support and a confusing message. A few random examples: PodcastAddict claims it supports "transcripts", but it only supports SRT files. Some apps claim they support cross-app comments, but they only support "activitypub" flavour cross-app comments. Podverse claims support for "value" but only supports boosts, not streaming sats. The specifications are too wide, in my view, and confuse "here's how we might use this specification for XXXX technology" with full support.

Personally, I would argue that Nano should be removed from the _specification_ , but it's absolutely fine to add Nano, Dogecoin, Hive, USD, darsek and the former Italian Lira as _optional future suggestions_ which at some point we agree to include in the specification.

So - we have a baseline of mandatory support for lightning (both streaming *and* boosts) in `podcast:value`, because it's in the specification. If you do not support lightning (both streaming **and** boosts), then you may not claim support for this tag. A specification is not a thing that we should encourage picking and choosing from. It's an absolute, and you cannot claim to be "Podcasting 2.0 Certified" if you do not correctly support a tag **in full**.

Accordingly, **nano should not be in this specification,** unless we all propose that it is also mandatorily supported by all apps. I do not believe that we have that agreement, and support its removal (or rewording as an _optional future suggestion_).

daveajones:
> Chiming in: a general complaint of mine about the podcast namespace is that there are many different things in the specification, and many aren't supported. This leads to patchy support and a confusing message. A few random examples: PodcastAddict claims it supports "transcripts", but it only supports SRT files. Some apps claim they support cross-app comments, but they only support "activitypub" flavour cross-app comments. Podverse claims support for "value" but only supports boosts, not streaming sats. The specifications are too wide, in my view, and confuse "here's how we might use this specification for XXXX technology" with full support.
> 
> Personally, I would argue that Nano should be removed from the _specification_ , but it's absolutely fine to add Nano, Dogecoin, Hive, USD, darsek and the former Italian Lira as _optional future suggestions_ which at some point we agree to include in the specification.
> 
> So - we have a baseline of mandatory support for lightning (both streaming _and_ boosts) in `podcast:value`, because it's in the specification. If you do not support lightning (both streaming **and** boosts), then you may not claim support for this tag. A specification is not a thing that we should encourage picking and choosing from. It's an absolute, and you cannot claim to be "Podcasting 2.0 Certified" if you do not correctly support a tag **in full**.
> 
> Accordingly, **nano should not be in this specification,** unless we all propose that it is also mandatorily supported by all apps. I do not believe that we have that agreement, and support its removal (or rewording as an _optional future suggestion_).

That's a convincing argument.

daveajones:
I’m going to merge this and then create a new file as a sort of alt coin registry to show a few examples.

brianoflondon:
I'm late to this but @jamescridland nailed it.

PR #390, opened 2022-09-03
Update 1.0.md

jamescridland:
Clarification about where to use channel and item for the podcast:person tag, by using minor celebrities from 1970s British television.


jamescridland:
Just a little nudge, @daveajones - was this OK?

daveajones:
Thanks for the bump.

PR #403, opened 2022-10-15
Use proper mimetypes

joksas:
- There were some old occurrences of `application/srt`, which I replaced with `application/x-subrip`
- Although `audio/mp3` can be found in some projects (including [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=227004)!), it is not recognised by IANA. `audio/mpeg` should be used instead.

### DIFF
--- a/transcripts/transcripts.md
+++ b/transcripts/transcripts.md
@@ -2,8 +2,12 @@
 
 This is the initial spec for the podcast transcript format. There are four possible formats detailed below.
 
-* Some transcript implementations are done in-browser. [CORS headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) are required to make these files available from other websites. [A CORS tester is available here](https://cors-test.codehappy.dev/), to ensure that transcripts are available within browser-based players.
+Some transcript implementations are done in-browser. [CORS headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) 
+are required to make these files available from other websites. [A CORS tester is available here](https://cors-test.codehappy.dev/), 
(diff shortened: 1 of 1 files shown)


jamescridland:
This looks like good work. I'd welcome this to be corrected.

jamescridland:
@daveajones would appreciate review

tomrossi7:
@joksas why the change to  `application/srt`?

joksas:
@tomrossi7, the change is from `application/srt` to `application/x-subrip`. Unfortunately, the former is not a [valid IANA mimetype](https://www.iana.org/assignments/media-types/media-types.xhtml), so [a breaking change](https://github.com/Podcastindex-org/podcast-namespace/pull/331) had to be introduced to fix the specification (names with `x-` prefix are ok for unregistered types).

We are in a tricky situation right now because although `application/x-subrip` is technically the correct mimetype to use, it might still be the case that many players haven't reflected this change and only support `application/srt`. I have personally contacted the developers behind Podverse and Podcast Addict, and they now display both mimetypes correctly. For *now*, I guess it is up to podcast hosting companies to decide which mimetype to use. At RSS Blue, we use `application/x-subrip`.

joksas:
I notice that @jamescridland is using both in [Podnews Daily RSS feed](https://podnews.net/rss). Has that worked the best for you, James?

jamescridland:
I figured it wouldn't do any damage to list both in the feed; it'll gzip down to virtually no additional bandwidth anyway.

But, while our own specs are incorrect, we'll see more of this issue unless it's fixed.

PR #405, opened 2022-10-20
Use "true"/"false" for `itunes:explicit`

joksas:
`<itunes:explicit>` uses "true"/"false" instead instead of "yes"/"no" for boolean value.

### DIFF
--- a/example.xml
+++ b/example.xml
@@ -32,7 +32,7 @@
         <podcast:medium>podcast</podcast:medium>
 
         <itunes:author>John Doe</itunes:author>
-        <itunes:explicit>no</itunes:explicit>
+        <itunes:explicit>false</itunes:explicit>
         <itunes:type>episodic</itunes:type>
         <itunes:category text="News"/>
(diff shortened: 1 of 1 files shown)


theDanielJLewis:
Apple used to accept yes/no on this tag in addition to true/false.

joksas:
What about now? Because [this webpage](https://help.apple.com/itc/podcasts_connect/#/itcb54353390) only list "true"/"false".

theDanielJLewis:
Sorry, I neglected to add that I agree with your clarification. I only meant to add the insight on what Apple _used_ to allow, but you're right that true/false is their spec now. (Though I suspect they still properly read yes/no.)

jamescridland:
It would have been helpful to see the source spec from Apple in this PR; probably why it's stalled. [It's over here](https://help.apple.com/itc/podcasts_connect/#/itcb54353390) and says, indeed, `true` or `false`

Given that this change is factually correct according to Apple's specification. I will merge this.

Issue #132, opened 2020-11-26
<podcast:monetization> Create tag to support Web Monetization on podcasts apps


benjaminbellamy:
I would like to suggest the creation of the \<podcast:monetization> tag in order to support Web Monetization (see https://webmonetization.org/ ).

Web Monetization just requires an Interledger Payment Pointer (like a wallet address) and that’s it!

Services such as https://uphold.com/ provide one in many currencies (you may choose USD, Euro, BitCoin…)
Services such as https://coil.com/ allow users to pay seamlessly. (Coil is $5/mo, no more, no less. $0.0001/second is sent to the podcaster.)

So you can get paid with "real" money.

So far Castopod is Web Monetization compliant (when users go to the web site).

Adding \<podcast:monetization> to the RSS feed would allow Web Monetization to work on podcasts apps.

The \<podcast:value> value is not compatible with Web Monetization because "method" and "suggested" are mandatory and requires elements.
The two systems (Lightning and Web Monetization) are really different so having different tags make sense.

For a web page, the tag is \<meta name="monetization" content="Interledger Payment Pointer"> so we would keep it alike.

I suggest the following:

- **\<podcast:monetization>[one "Interledger Payment Pointer" element]</podcast:monetization>**

Channel

(optional | single)

This element defines the Interledger Payment Pointer for Web Monetization.

jamescridland:
I'd not heard of this before - thanks so much for sharing. I've just added the monetization tag to all Podnews network websites. See it, for example, in https://pod.events 

I'd *really* like to see whether we can get this to work somehow with an existing funding tag. Might I suggest...

`<podcast:value type="webmonetization">[one "Interledger Payment Pointer" element]</podcast:value>`

or

`<podcast:funding type="webmonetization" pointer="[one "Interledger Payment Pointer" element]"></podcast:funding>`

The latter would make more sense, perhaps. In this example, it doesn't have a `url` attribute, so the UI knows that there's no button and no link visible.

I'm already a bit grumpy of the existence of :funding and :value without the addition of :monetization - do you think the current ones are extendable?

brianoflondon:
I think the ```<value>``` tag has more than enough scope for this. 

Even though there are no clients or infrastructure to support this yet, my feed has the following in it for distributing Hive Blockchain Dollars (a USD linked stable coin within the Hive blockchain). My address is simply [@brianoflondon](https://peakd.com/@brianoflondon/) which makes things very readable:
```
[...]

PR #409, opened 2022-10-29
Proposal for webmonetization using value

kookster:
Follow up to: https://github.com/Podcastindex-org/podcast-namespace/issues/132

### DIFF
--- a/proposal-docs/value-webmonetization/value-webmonetization.md
+++ b/proposal-docs/value-webmonetization/value-webmonetization.md
@@ -1,0 +1,63 @@
+# Web Monetization using `<podcast:value/>`
+
+To enable and promote Web Monetization (WM) in podcasting, necessary information for WM could be provided for a podcast by adding a `<podcast:value/>` tag to the RSS feed.
+  https://github.com/Podcastindex-org/podcast-namespace/blob/main/value/value.md
+
+This `value` tag would have the following values for use with Web Monetization:
+```xml
+<podcast:value 
+    type="webmonetization" 
+    method="ILP"
+></podcast:value>
+```
+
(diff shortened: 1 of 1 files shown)


proposal-docs/value-webmonetization/value-webmonetization.md
```diff
@@ -0,0 +1,63 @@
...
+    <podcast:valueRecipient
+        name="Alice"
+        type="paymentpointer"
+        address="$example.now/~alice"
+        split="100"
+    >
+</podcast:value>
+```
+
+While payments in WM are streamed to a single payment pointer at a time, the WM authors suggest splits can be implemented using [Probabilistic Revenue Sharing](https://webmonetization.org/docs/probabilistic-rev-sharing) or perhaps for WM should be [implemented at the payment pointer](https://github.com/Podcastindex-org/podcast-namespace/issues/132#issuecomment-780145285) rather than expecting the client to be responsible.  In either case, the `podcast:value` and `podcast:valueRecipient` tags can provide sufficient information for a web player or page to setup Web Monetization for the podcast.
```
yassinedoghri, line 36: Not sure about the suggestion, but the second part of the first sentence seemed off to me.

```suggestion
While payments in WM are streamed to a single payment pointer at a time, the WM authors suggest splits can be implemented using [Probabilistic Revenue Sharing](https://webmonetization.org/docs/probabilistic-rev-sharing) or perhaps it should be [implemented at the payment pointer](https://github.com/Podcastindex-org/podcast-namespace/issues/132#issuecomment-780145285) rather than expecting the client to be responsible. In either case, the `podcast:value` and `podcast:valueRecipient` tags can provide sufficient information for a web player or page to setup Web Monetization for the podcast.
```

PR #412, opened 2022-11-07
Update GUID

jamescridland:
Adding a link to the tool from RSS Blue


jamescridland:
Since there has been no objection, I'm merging this PR.

PR #414, opened 2022-12-04
Do not use HTTP protocol for URLs in podcast namespace tags

joksas:
This is to comply with the following in the [podcast namespace specification](https://podcastindex.org/namespace/1.0#podcast-tags):

> Anywhere the url of a hyper-text based resource is specified, it must be given as `https:` and not `http:`.

### DIFF
--- a/docs/1.0.md
+++ b/docs/1.0.md
@@ -240,19 +240,19 @@ The full taxonomy list is [here](https://github.com/Podcastindex-org/podcast-nam
 
 ### Examples
 ```xml
-<podcast:person href="https://example.com/johnsmith/blog" img="http://example.com/images/johnsmith.jpg">John Smith</podcast:person>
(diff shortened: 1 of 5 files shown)


daveajones:
Thanks!  Will review/merge today.

jamescridland:
I'd like to query the requirement here. It is best practice to use `https://` for everything, but I don't see why `http://` is not acceptable. Is there a technical reason?

theDanielJLewis:
I saw this update from Brave Browser that covers a little bit about the complications of HTTPS default and some things to consider: https://brave.com/privacy-updates/22-https-by-default/

brianoflondon:
Unfortunately https seems to be the only way to go. As soon as you try to mix https and http all the browsers scream at you.

For example many new TLDs (like .app which I'm using) is set to ONLY work on https. That is set by default.

PR #417, opened 2022-12-05
socialInteract: URI vs URL text

jamescridland:
Clarifying which to use in the URI field if you're given both a URI and URL, as boringly hammered out in podcast.social over the weekend.


jamescridland:
Bumping this for review

jamescridland:
Given this was agreed in the social discussion, and that there are no objections to it, I'm taking the decision to merge this PR.

Issue #428, opened 2023-01-05
Why only one Value block per RSS element at each level?


brianoflondon:
Could anyone explain to me WHY we only allow one Value element at each level? I hadn't even been aware that this got set this way but I've had Hive as a value element in all 3speak RSS feeds for as long as I can remember (which still constitute by far the greatest number of value enabled feeds). 

As far as I know having the Hive value block (whilst completely useless for now) doesn't break anybody's apps. 

I'm asking this because I am talking with the person who is making a 3speak App/web app for mobile about putting in an RSS reader and Value 4 Value, and it would be easy for him to do streaming native Hive payments to all my 3speak creators and Lightning to the others.

https://github.com/Podcastindex-org/podcast-namespace/blob/main/value/value.md

> Value Element
>The <podcast:value> tag designates the cryptocurrency or payment layer that will be used, the transport method for transacting the payments, and a suggested amount denominated in the given cryptocurrency.
>
>This element can exist at either the <channel> or <item> level.
>
>**Currently, there can be only a single copy of this element at each level.**

Current 3speak RSS:

```rss
        <podcast:value type="lightning" method="keysend" suggested="0.00000050000">
            <podcast:valueRecipient name="Brian of London" address="0266ad2656c7a19a219d37e82b280046660f4d7f3ae0c00b64a1629de4ea567668" customKey="818818" customValue="brianoflondon" type="node" split="99"></podcast:valueRecipient>
            <podcast:valueRecipient name="PodcastIndex" address="03ae9f91a0cb8ff43840e3c322c4c61f019d8c1c3cea15a25cfc425ac605e61a4a" type="node" fee="True" split="1"></podcast:valueRecipient>
        </podcast:value>
        <podcast:value type="HBD" method="transfer" suggested="0.05">
            <podcast:valueRecipient name="podcaster" type="account" address="brianoflondon" split="98"></podcast:valueRecipient>
            <podcast:valueRecipient name="host" type="account" address="threespeak" split="1"></podcast:valueRecipient>
            <podcast:valueRecipient name="podcastindex" type="account" address="podcastindex" split="1"></podcast:valueRecipient>
        </podcast:value>
```

jamescridland:
+1

I accept VISA, MasterCard and American Express - whatever the customer wants. The same should work here.

Podnews Daily's current RSS:

```xml
<podcast:value type="lightning" method="keysend" suggested="0.00000015000">
[...]

PR #429, opened 2023-01-05
Remove requirement for only one Value block

brianoflondon:
As per the discussion: https://github.com/Podcastindex-org/podcast-namespace/issues/428

#428


brianoflondon:
@daveajones you can merge this if we're happy with removing the suggestion of only one value block.

daveajones:
#132

joksas:
<https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/1.0.md#value> should also be updated, right? Because it currently says that count is single.

PR #438, opened 2023-02-11
Fix a typo in the spec

MikeKemmerer:


jamescridland:
Given this is a simple typo, I will merge.

Issue #351, opened 2022-03-14
Proposal - podcast:complete tag


daveajones:
As we move forward with slowly providing replacement tags for itunes namespace tags, I had a thought:  in replicating the `<itunes:complete>` tag, why not add an optional link to a full archive feed if it was a long running podcast and had many hundreds of episodes.  The main feed might not want to carry all of those episodes.  But, an archive link to a full feed would be nice.

### **\<podcast:complete>**

<br>

<b>

```xml
<podcast:complete
 archive="[feed url(string)]"
>
[yes|no(bool)]
</podcast:complete>
```

</b>

#### (optional | single)

This element allows a podcaster to signal to the world that this podcast will never publish another episode to this feed.  In it's basic form, it is a direct drop-in
replacement for `<itunes:complete>` and functions identically.  The addition of the `archive` attribute allows for specifying the url of a complete archival feed of the
podcast that contains every episode in case this feed does not.

Specifying a value of `no` is the same as this tag not being present.

#### Attributes
 - **archive** (optional) A url of a feed that contains every episode of this podcast.

Examples:
```xml
<!-- This podcast will never publish again -->
<podcast:complete>yes</podcast:complete>

<!-- Redundant usage and not recommended -->
<podcast:complete>no</podcast:complete>

<!-- This podcast is over, and a complete episode archive feed is "here" -->
<podcast:complete archive="https://example.org/rss/myarchive.xml">yes</podcast:complete>
```

theDanielJLewis:
I'm fine with replicating the tag, but I question the necessity of providing an archive feed when the master feed itself could contain all episodes.

Under what scenario would someone want to mark a feed complete, _not_ let it include all the episodes, but still link to a feed that _does_ have all episodes?

I can see people wanting to "complete" an entire show with all episodes.

I can also see people wanting to "complete" a show but keep only the latest episodes available.

I just can't see why they would want a hybrid.

evoterra:
Agree with DJL on the archive link. Though the opposite is interesting. To me, and archive feed _is_ a complete feed. So if anything, and archive/complete feed should point back to a _still being updated_ feed. 

Use case: A 10-episode series on the history of prohibition that’s _complete_, but the host has an ongoing history podcast.

theDanielJLewis:
Good to see you here, @evoterra! (Or am I forgetting that you've already been here for a while?)

I like that idea for sure! And I think it would be better for the related/featured tag discussed elsewhere.

[...]

PR #439, opened 2023-02-14
Proposal for <podcast:updateFrequency>

nathangathright:
Following up on the discussion around [the podcast:complete tag](https://github.com/Podcastindex-org/podcast-namespace/issues/351), here’s my proposal for an expanded use case.

# The `<podcast:updateFrequency>` Specification

<small>Version 1.0 by [Nathan Gathright](https://github.com/nathangathright) - 2023.02.13</small>

> A tag to provide podcasters a way to express their intended schedule for future releases.

## Purpose
In the spirit of taking platform-specific innovations and making them accessible to the open ecosystem of RSS, the intent of this proposal is to replicate the purpose of the “Update Frequency” field in Apple Podcasts Connect as well as replace the `<itunes:complete>` tag for podcasts that have no intention to release further episodes.

Additionally, apps like Overcast, Pocket Casts, and Podchaser analyze the intervals between past episodes to estimate a prediction for future releases. Under certain conditions or naive algorithms, this can be inaccurate:
* New podcasts that only have a trailer
* Limited-run podcasts nearing the end of a season
* “Daily” podcasts that only release on weekdays
* Podcasts with consistent releases at irregular intervals (every Monday and Wednesday)

If podcasters can unambiguously communicate their release schedule, then more apps can provide accurate information to their listeners.

## Parent
`<channel>`

## Count
Single

## Node value

The node value is a free-form string, which might be displayed alongside other information about the podcast. Please do not exceed `128 characters` for the node value or it may be truncated by aggregators.

## Attributes
* **complete (optional):** Boolean specifying if the podcast has no intention to release further episodes. If not set, this should be assumed to be false.
* **dtstart (optional):** The `date` or `datetime` the recurrence rule begins. If the `rrule` contains a value for `COUNT`, then this attribute is required. If the `rrule` contains a value for `UNTIL`, then the value of this attribute must be formatted to the same date/datetime standard.
* **rrule (recommended):** A recurrence rule as defined in [iCalendar RFC 5545 Section 3.3.10](https://www.rfc-editor.org/rfc/rfc5545#section-3.3.10). Podcasters should not be expected to be able to write a valid recurrence rule themselves. There are [several libraries](https://github.com/topics/rrule) to generate valid recurrence rules from form data or natural language text inputs.

## Examples

Recreating most of Apple Podcasts Connect’s “Update Frequency” values is easily achieved:
```xml
<podcast:updateFrequency rrule="FREQ=DAILY">Daily</podcast:updateFrequency>
<podcast:updateFrequency rrule="FREQ=WEEKLY">Weekly</podcast:updateFrequency>
<podcast:updateFrequency rrule="FREQ=WEEKLY;INTERVAL=2">Biweekly</podcast:updateFrequency>
<podcast:updateFrequency rrule="FREQ=MONTHLY">Monthly</podcast:updateFrequency>
<podcast:updateFrequency rrule="FREQ=MONTHLY;INTERVAL=2">Bimonthly</podcast:updateFrequency>
<podcast:updateFrequency rrule="FREQ=MONTHLY">Monthly</podcast:updateFrequency>
<podcast:updateFrequency rrule="FREQ=YEARLY">Yearly</podcast:updateFrequency>
```

However, greater precision can be easily communicated:
```xml
<podcast:updateFrequency rrule="FREQ=DAILY;INTERVAL=1;BYDAY=MO,TU,WE,TH,FR">Every weekday</podcast:updateFrequency>
<podcast:updateFrequency rrule="FREQ=WEEKLY;BYDAY=MO,WE">Every Monday and Wednesday</podcast:updateFrequency>
<podcast:updateFrequency rrule="FREQ=WEEKLY;BYDAY=FR;BYMONTHDAY=13">Every friday the 13th</podcast:updateFrequency>
<podcast:updateFrequency rrule="FREQ=YEARLY;BYDAY=+4TH;BYMONTH=11">Every year on American Thanksgiving</podcast:updateFrequency>
```

Limited-run series can set expectations clearly:
```xml
<podcast:updateFrequency rrule="FREQ=WEEKLY;INTERVAL=2;BYDAY=MO;COUNT=10" dtstart="20230828">Every other Monday for 10 episodes starting on Aug 28, 2023</podcast:updateFrequency>
<podcast:updateFrequency rrule="FREQ=WEEKLY;UNTIL=20231231;BYDAY=MO">Every Monday until Dec 31, 2023</podcast:updateFrequency>
```

Completed series with no intention to release further episodes:
```xml
<podcast:updateFrequency completed="true">That’s all folks!</podcast:updateFrequency>
```

### DIFF
--- a/proposal-docs/updateFrequency/updateFrequency.md
+++ b/proposal-docs/updateFrequency/updateFrequency.md
@@ -27,7 +27,8 @@ The node value is a free-form string, which might be displayed alongside other i
 
 ## Attributes
 * **complete (optional):** Boolean specifying if the podcast has no intention to release further episodes. If not set, this should be assumed to be false.
-* **rrule (optional):** A recurrence rule as defined in [iCalendar RFC 5545 Section 3.3.10](https://www.rfc-editor.org/rfc/rfc5545#section-3.3.10). Podcasters should not be expected to be able to write a valid recurrence rule themselves. There are [several libraries](https://github.com/topics/rrule) to generate valid recurrence rules from form data or natural language text inputs.
+* **dtstart (optional):** The `date` or `datetime` the recurrence rule begins. If the `rrule` contains a value for `COUNT`, then this attribute is required. If the `rrule` contains a value for `UNTIL`, then the value of this attribute must be formatted to the same date/datetime standard.
+* **rrule (recommended):** A recurrence rule as defined in [iCalendar RFC 5545 Section 3.3.10](https://www.rfc-editor.org/rfc/rfc5545#section-3.3.10). Podcasters should not be expected to be able to write a valid recurrence rule themselves. There are [several libraries](https://github.com/topics/rrule) to generate valid recurrence rules from form data or natural language text inputs.
 
 ## Examples
 
@@ -44,7 +45,7 @@ Recreating most of Apple Podcasts Connect’s “Update Frequency” values is e
 
 However, greater precision can be easily communicated:
 ```xml
-<podcast:updateFrequency rrule="FREQ=DAILY;INTERVAL=1;WKST=MO;BYDAY=MO,TU,WE,TH,FR">Every weekday</podcast:updateFrequency>
+<podcast:updateFrequency rrule="FREQ=DAILY;INTERVAL=1;BYDAY=MO,TU,WE,TH,FR">Every weekday</podcast:updateFrequency>
 <podcast:updateFrequency rrule="FREQ=WEEKLY;BYDAY=MO,WE">Every Monday and Wednesday</podcast:updateFrequency>
 <podcast:updateFrequency rrule="FREQ=WEEKLY;BYDAY=FR;BYMONTHDAY=13">Every friday the 13th</podcast:updateFrequency>
 <podcast:updateFrequency rrule="FREQ=YEARLY;BYDAY=+4TH;BYMONTH=11">Every year on American Thanksgiving</podcast:updateFrequency>
@@ -52,7 +53,7 @@ However, greater precision can be easily communicated:
 
 Limited-run series can set expectations clearly:
 ```xml
-<podcast:updateFrequency rrule="FREQ=WEEKLY;INTERVAL=2;COUNT=10">Every other week for 10 episodes</podcast:updateFrequency>
+<podcast:updateFrequency rrule="FREQ=WEEKLY;INTERVAL=2;BYDAY=MO;COUNT=10" dtstart="20230828">Every other Monday for 10 episodes starting on Aug 28, 2023</podcast:updateFrequency>
 <podcast:updateFrequency rrule="FREQ=WEEKLY;UNTIL=20231231;BYDAY=MO">Every Monday until Dec 31, 2023</podcast:updateFrequency>
 ```
 
--- a/proposal-docs/updateFrequency/updateFrequency.md
+++ b/proposal-docs/updateFrequency/updateFrequency.md
@@ -1,0 +1,62 @@
+# The `<podcast:updateFrequency>` Specification
+
+<small>Version 1.0 by [Nathan Gathright](https://github.com/nathangathright) - 2023.02.13</small>
+
+> A tag to provide podcasters a way to express their intended schedule for future releases.
+
+## Purpose
+In the spirit of taking platform-specific innovations and making them accessible to the open ecosystem of RSS, the intent of this proposal is to replicate the purpose of the “Update Frequency” field in Apple Podcasts Connect as well as replace the `<itunes:complete>` tag for podcasts that have no intention to release further episodes.
+
+Additionally, apps like Overcast, Pocket Casts, and Podchaser analyze the intervals between past episodes to estimate a prediction for future releases. Under certain conditions or naive algorithms, this can be inaccurate:
+* New podcasts that only have a trailer
+* Limited-run podcasts nearing the end of a season
+* “Daily” podcasts that only release on weekdays
+* Podcasts with consistent releases at irregular intervals (every Monday and Wednesday)
+
+If podcasters can unambiguously communicate their release schedule, then more apps can provide accurate information to their listeners.
+
+## Parent
+`<channel>`
+
+## Count
+Single
+
+## Node value
+
+The node value is a free-form string, which might be displayed alongside other information about the podcast. Please do not exceed `128 characters` for the node value or it may be truncated by aggregators.
+
+## Attributes
+* **complete (optional):** Boolean specifying if the podcast has no intention to release further episodes. If not set, this should be assumed to be false.
+* **rrule (optional):** A recurrence rule as defined in [iCalendar RFC 5545 Section 3.3.10](https://www.rfc-editor.org/rfc/rfc5545#section-3.3.10). Podcasters should not be expected to be able to write a valid recurrence rule themselves. There are [several libraries](https://github.com/topics/rrule) to generate valid recurrence rules from form data or natural language text inputs.
+
+## Examples
+
+Recreating most of Apple Podcasts Connect’s “Update Frequency” values is easily achieved:
+```xml
+<podcast:updateFrequency rrule="FREQ=DAILY">Daily</podcast:updateFrequency>
+<podcast:updateFrequency rrule="FREQ=WEEKLY">Weekly</podcast:updateFrequency>
+<podcast:updateFrequency rrule="FREQ=WEEKLY;INTERVAL=2">Biweekly</podcast:updateFrequency>
+<podcast:updateFrequency rrule="FREQ=MONTHLY">Monthly</podcast:updateFrequency>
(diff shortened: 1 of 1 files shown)


jamescridland:
(I added the proposal to the first message so others can comment more easily)

I really like the use of `rrule`. It looks really useful to me, and I wonder if it should be "recommended" rather than "optional".

brianoflondon:
As nice as this is, for this to be any use it would once again fall to podcast hosting companies to implement and indexing sites like PodcatsIndex to check and take action on.

In terms of work to add to hundreds of rss feeds, including adding a UI so podcasters can set this, it would seem to be lots more work than integrating with Podping which makes this irrelevant and avoids the need for someone to manually set this value.

nathangathright:
@brianoflondon maybe I’m missing something about podping. It was my understanding that it improved notification of new episodes, nothing to do with communicating a podcaster’s intended schedule for future releases to their listeners. So, I’m unclear how it makes this irrelevant.

brianoflondon:
If the goal is communication with end users, I suspect that saying "Our show is daily at around 8am EST" or "We drop new shows on Thursdays and Sunday every week" in the description is going to help more users than a machine readable format which will need both hosts and clients to implement.

If the goal is something which can be acted on programmatically i.e. to pick up new episodes, I go back to pointing out that Podping does this better and is already working.

theDanielJLewis:
Why `updateFrequency` instead of simply `frequency`?

I would suggest offering a predefined list of node values that could be suggested and chosen, while still allowing a custom entry.

- Daily
- Semiweekly
- Weekly
- Biweekly
- Semimonthly
- Monthly
- Yearly
- Completed

And relevant to @evoterra's observations, maybe there should be an extra attribute like `seasonal="true"` so a show could indicate it releases weekly _during its active season_.

@brianoflondon, I think Marco would love to see this in the feed instead of making his system try to determine it. Plus, it would be worth asking him _why_ he displays the frequency to his users and what their responses have been to that feature.

francosolerio:
Podping is already working for apps that have a server component that
implements it.  There's no way at this moment for apps with a server-less
stack to receive podping notifications. Having the frequency, or even
better the usual publication approx  time, would be very helpful to spare
bandwidth and battery.

F

brianoflondon:
Marco seems to have solved this. I had a very quick try with some python looking at podping data but came to realise it's really not easy. I'm not sure Marco would like us stomping on one of his differentiators though I can't think that I've missed that info since leaving Overcast. 

I still can't see this getting any widespread adoption on either side of publishers and consumers.

ericpp:
> Why `updateFrequency` instead of simply `frequency`?

Or `schedule` or `releaseSchedule` if it's denoting the podcast release schedule.

nathangathright:
@evoterra 

1. A serialized show can start with an indefinite recurrence rule and update it once they’ve schedule the end of the current season. They don’t need to have everything figured out from the jump.
2. For seasonal shows, the update frequency should only represent the intended release schedule for upcoming episodes. I don’t see the value of including release frequency for already-released content. They’re already out. You can binge them all.  Help me understand if I’m missing something.
3. To indicate that a season is complete but the show will continue seems like a better fit as an attribute on [the season tag](https://github.com/Podcastindex-org/podcast-namespace/issues/351#issuecomment-1309136640).
4. If a seasonal show isn’t currently releasing episodes, but plans to in the future, then maybe we need to add a `startdate` attribute to the tag. (This actually addresses multiple concerns.)

evoterra:
@nathangathright 

1. Good
2. Agree the prior release schedule is meaningless for complete seasons. It'll be on the hosting providers (who'll cater to this type of content) to get the latest release schedule from providers. That makes sense.
3. Agree that we need to extend the season tag and not conflate here.
4. That's a good idea. And, as mentioned in #2, it's something the hosting provider can help remind the creator to set. (How is TBT, obvs.)

nathangathright:
@brianoflondon I agree Podping is the better tool for picking up new episodes. The goal here is to achieve feature parity in the namespace with Apple Podcasts, and they wanted it in their UI badly enough to add a field to Apple Podcasts Connect rather than wait for all the hosts to support a new `<itunes>` tag.

It’s far from a differentiator for Marco when other third-party apps are also supporting it.
Frame 17

If fact, I wrote my own frequency predictor for an older version of Podlink. All of my example edge cases in this proposal are based on real-world observations where an app’s look-back analysis was too naive or otherwise incapable of recognizing the pattern that could be clearly communicated by the podcaster. Just putting it in the description doesn’t enable apps to reach feature parity with Apple Podcasts.

nathangathright:
> > Why `updateFrequency` instead of simply `frequency`?
> 
> Or `schedule` or `releaseSchedule` if it's denoting the podcast release schedule.

@theDanielJLewis @ericpp The explicit purpose of this tag is to achieve feature parity in the namespace with the field in Apple Podcasts Connect, and "frequency" is too confusing given that we’re talking about audio.

To quote Dave Winer’s [Rules For Standards, Perfection is a waste of time](http://scripting.com/2017/05/09/rulesForStandardsmakers.html#:~:text=Perfection%20is%20a%20waste%20of%20time):
> It totally doesn't matter what we call it. We can learn to use anything. There are more important things to spend time on.

nathangathright:
> I would suggest offering a predefined list of node values that could be suggested and chosen, while still allowing a custom entry.
@theDanielJLewis If a host wants to display a predefined list of node values, that’s their prerogative. If you’re not objecting to custom text entry, then it makes no difference to the spec.

> maybe there should be an extra attribute like `seasonal="true"` so a show could indicate it releases weekly _during its active season_.
After thinking through Evo’s concerns, I think a startdate attribute might address a few various concerns. An inactive show should specify when they plan to resume their weekly schedule or remove the update frequency from their feed.

jamescridland:
> The explicit purpose of this tag is to achieve feature parity in the namespace with the field in Apple Podcasts Connect

Our purpose is to **add new features that benefit podcast listeners and creators**. It isn't anything else.

I wonder why Apple put this data in Podcasts Connect, rather than add a field in the RSS for it?

I also wonder how much full of crap the data is, given that this will be added when you launch a podcast and probably will never be touched by a podcaster after launch. It strikes me that this is best to be algorithmically worked out by looking at the RSS feed. I'd really like to do some work on comparing update frequency in Apple Podcasts Connect with reality.

Evo's comments are interesting. Instead of an update frequency, are we better having a tag which is "future episodes", which allows us to suggest that "there will be six episodes releasing weekly on a Thursday from 21 April 2023?" There is a recurring calendar specification for this, which seems quite robust.

nathangathright:
> I also wonder how much full of crap the data is, given that this will be added when you launch a podcast and probably will never be touched by a podcaster after launch. It strikes me that this is best to be algorithmically worked out by looking at the RSS feed.

When podcasters don’t maintain their intended schedule is precisely when these algorithms fail. They’re not robust enough to account for the variance, or there’s not enough data to make it obvious what’s an outlier versus a change in the publishing schedule. Personally, I’d prefer to know the podcaster’s intentions, even if they don’t live up to them. If apps want to corroborate an update frequency before showing it their listeners, I hope the `rrule` make that task easier.

I’ve strived to provide specific examples where the existing algorithms fail to grasp reasonable release schedules. If you check Pocket Casts on a Friday or Saturday, their page for Podnews will confidently you, "Daily podcast. Next episode tomorrow." In this case, their _best_ isn’t accurate. 

> I'd really like to do some work on comparing update frequency in Apple Podcasts Connect with reality.

In July 2021 (3 months after the feature launched), Pacific Content found that [only 14% of all shows in Apple Podcasts have set an update frequency](https://blog.pacific-content.com/how-often-are-most-podcasts-updated-f187b8cc1931#:~:text=weeks.%20So%20far%2C-,about,-14%25%20of%20all). We’re only about 2 months away from the 2-year anniversary of the feature.

evoterra:
Algorithms will fail at this. Massively. Sure, it's fine for a show like Podnews or Geek News Central, which puts out content on a consistent basis. 

But, and I know I'm being a broken record but you also know I don't give two shits about being called out for being a broken record, the algorithm cannot handle serialized content. Nor can it handle when extended breaks are taken and episodes resume again. Not with any accuracy. And, to @nathangathright's point, the show's intention the frequency of  future releases are completely unknown by any algorithm. (Not that I would be opposed to an algorithm that triggered when a show fails to live up to that expectation, mind you. No reason we can't have both. But that algorithmic trigger better be solid.)

nathangathright:
To better visualize the proposal, here’s a sample UI implementation to show how a host might implement this in their own dashboards:
* Demo: https://nathangathright.github.io/updateFrequency/
* Source code: https://github.com/nathangathright/updateFrequency

theDanielJLewis:
On the intention-versus-reality side, publishing tools could auto-populate the tag in accordance with past episodes. So if a podcaster marked it as "weekly," but is publishing inconsistently, the publishing tool could (either itself or by prompt) switch to "not set schedule."

PR #510, opened 2023-04-20
Include the bLIP 10 Specification in this Repository

Alwin-Stockinger:
This PR addresses [https://github.com/Podcastindex-org/podcast-namespace/discussions/506](https://github.com/Podcastindex-org/podcast-namespace/discussions/506).

It also includes the action proposed in [https://github.com/Podcastindex-org/podcast-namespace/discussions/500](https://github.com/Podcastindex-org/podcast-namespace/discussions/500).

## Changes to the [original document](https://github.com/lightning/blips/blob/master/blip-0010.md):

- Removed the blip number, title and status from the heading and changed Author to Original Author
- Modified the abstract to better fit with the rest of Podcasting 2.0
- Turned codeblocks into json codeblocks
- Corrected the Example in the Abstract to a valid json with double quotes instead of single quotes
- Made the “Formatting and encoding” section a little clearer
- Made clear that multiple fields for identifying podcast and episode can be added
- Added guid to “use any of `podcast`, `feedID` or `url`”
- Made `episode_guid` preferred field for identifying episode because `episode_guid` does not rely on the episode being in the index
- Put `ts`above `time` to reflect that ts is preferred
- Add `auto` option to action, as was proposed by @dhk2 in [https://github.com/Podcastindex-org/podcast-namespace/discussions/500](https://github.com/Podcastindex-org/podcast-namespace/discussions/500)
- Added to `boost_link` field description that it can be used for any action and that links may not contain podcast, episode AND timestamp, as Fountain currently only links to the episode not a timestamp.
- Switched recommended from `name` to `sender_name` because the receiver should know the name of their split anyway, but not the one of the sender
- Reformulated `sender_id` description
- Removed `pubkey`, `seconds_back`, `sender_key`, `sig_fields`, `signature` fields because no app is using them (Podverse sends the `pubkey` field but only puts the value “podverse-pubkey" in it)
- Removed `value_msat` because that information is in the invoice anyways
- Made `value_msat_total` **recommended** for numerology purposes
- Removed Motivation section
- Removed Rationale section
- Removed Universality section
- Removed Backwards Compatibility section
- Added Rust-V4V to Reference Implementations
- Added Examples & App Specifics section

### DIFF
--- a/value/blip-0010.md
+++ b/value/blip-0010.md
@@ -1,0 +1,265 @@
+# bLIP 10
+
+```text
+Original Author: Satoshis Stream <satoshisstream@pm.me>
+Created: 2022-01-24
+License: CC0
+```
+
+## Abstract
+
+Using Value 4 Value apps, listeners can send sats to the hosts of the podcasts they listen to. Per minute sending is called `stream` and manual payments are called `boost`s. If there is a message to the podcaster, with a boost, it is called a `boostagram`.
+These apps (see [Reference Implementations](#reference-implementations)) adopted bLIP 10 with the **7629169** TLV type for inputting key-value JSON metadata about the sent payment. The TLV holds data about the timestamp when the payment was sent within the episode, the name of the podcast and its [Guid](https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/1.0.md#guid), among other fields described in the [Specification](#specification) section.
+
+Example:
+
+```json
+{
+    "podcast": "PODCASTNAME",
+    "feedID": 1337,
+    "episode": "EPISODENAME",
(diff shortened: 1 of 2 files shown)


satoshisstream:
Good idea to remove the old/unused things.

satoshisstream:
https://github.com/lightning/blips/pull/23

Alwin-Stockinger:
> lightning/blips#23

Is there any automatic way we could keep both of them in sync somehow?

satoshisstream:
Just link it I'd say; anyone can PR the lightning repo :)

daveajones:
Thanks for this. Much easier to track now.

PR #530, opened 2023-05-10
add simplecast to txt element support

sophia-chapman:
Simplecast now supports the `<podcast:txt>` element and we would like to be added to the application list!


daveajones:
Thanks Sophia!

PR #552, opened 2023-07-10
Fixing broken markdown links

nathangathright:
The formatting introduced in 4967ffcc95a085767b431b72c5c8b6ecb5742b48 broke many markdown links. This commit removes only the line breaks that are breaking links as well converting some absolute links to relative.


daveajones:
Merged the one from @stevencrader.  Thanks Nathan.

PR #557, opened 2023-07-28
Update element-support.md

Agorise:
Added CastGarden (https://Cast.Garden)


Agorise:
Mobile, Hive, Web... It would be nice if the Fediverse platform could be added at some point as well. Syndicating your podcasts to the Fediverse as well can bring you thousands of additional listeners/viewers, ie: 
https://cast.garden/about/follows

PR #568, opened 2023-08-29
fix typo in fast-follow link example

Marzal:
Not sure though if the example is really valid as this part of the link is what podnews is using https://podnews.net/podcast/i8xe9

Also tried as an alternative without luck https://curiocaster.com/#fastfollow-podcast:396d9ae0-da7e-5557-b894-b606231fa3ea

### DIFF
--- a/docs/1.0.md
+++ b/docs/1.0.md
@@ -752,8 +752,7 @@ different platforms.
 
 &nbsp; `type` is currently `podcast`, but may be extended in future.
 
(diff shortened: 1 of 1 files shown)


jamescridland:
I'll merge this (there was a buggy pull request a while ago which added linebreaks all over the place)

I do wonder whether the fast-follow idea should be removed from the specs, given it isn't supported by anyone. (Podnews no longer supports it in those QR codes).

PR #587, opened 2023-11-25
Added note for bLIP 10 fields removed by #510

ericpp:


ericpp:
Since `value_msat` is still widely used, maybe it should be re-added to the spec? In theory, it should always match the number of received sats in the invoice.

johnspurlock:
`value_msat` would be very useful to include in the spec (it's in the example!), as it can contain millisats (i don't think invoices can or at least do?) - millisats are useful to back into the total amount if you have a small split

and it's a great indication of "what the client meant to send", to be sanity checked against what was actually received

also, as you say, already being sent by many

johnspurlock:
this was my only recommended note for whoever is in charge of the spec over in https://reflex.livewire.io/apps
image

ericpp:
@johnspurlock Invoices do contains both sat and msat totals:
e.g. lncli output
```
"amt_paid": "34000",
"amt_paid_sat": "34",
"amt_paid_msat": "34000",
...
"htlcs": [
    {
        ...
        "amt_msat": "34000",
    }
]
```

`value_msat` might be useful since it contains the original split value before any lightning routing fees are taken out. I think this can still be derived from the HTLCs in the  invoice, but it might be more convenient to get it from the TLV. On the otherhand, getting it from the invoice is better when reporting actual msat totals, since TLVs can be easily spoofed.

samsethi:
Thanks to @johnspurlock we added both value_msat and value_msat_total to the Alby Dashboard. Are we missing anything else John? 

Alby__Your_Bitcoin___Nostr_companion_for_the_web

PR #624, opened 2024-04-03
Updating the specs for WebVTT

tomrossi7:
Buzzsprout will be rolling out our support for VTT and I wanted to make sure the specs were up-to-date. I don't believe there are any character limits. I can't find any rhyme or reason to how Apple is doing it in their transcripts (other than speaker changes).

@jamescridland does this look right?

### DIFF
--- a/transcripts/transcripts.md
+++ b/transcripts/transcripts.md
@@ -162,34 +162,32 @@ Web Video Text Tracks Format (WebVTT) are an alternative to SRT primarily design
 - Cue settings allow the customization of cue positioning on the video
 
 #### Properties:
-- Max number of lines: 2
-- Max characters per line: 32
-- Speaker names (optional): Start a new card when the speaker changes. Include the speaker's name, followed by a colon.
+- Speaker names (optional): Speakers can be included in a voice span tag `<v>` at the beginning of each caption.
 
 #### Snippet:
 ```
 WEBVTT
 
-00:00:00.000 --> 00:00:02.760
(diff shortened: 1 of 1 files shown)


theDanielJLewis:
FYI, it seems Apple's downloadable transcripts break at sentences instead of shorter segments. I'm surprised they don't even break at punctuation to help shorten lines from long sentences.

Maybe they've studied and found it's actually easier to read the transcript—if the full transcript is shown—when a full sentence is highlighted at a time instead of smaller sections. That's certainly easier than word-by-word highlighting (which looks cool but isn't actually very readable).

tomrossi7:
> FYI, it seems Apple's downloadable transcripts break at sentences instead of shorter segments. I'm surprised they don't even break at punctuation to help shorten lines from long sentences.

Yeah, the standard doesn't really provide any guidelines for how long a caption can be. I would think shorter segments are better since you get a higher fidelity on your timestamps.

> Maybe they've studied and found it's actually easier to read the transcript—if the full transcript is shown—when a full sentence is highlighted at a time instead of smaller sections. That's certainly easier than word-by-word highlighting (which looks cool but isn't actually very readable).

I thought they were doing word-by-word highlighting. I wonder if they use the VTT as input and then have another algorithm for the highlighting?

theDanielJLewis:
> I thought they were doing word-by-word highlighting. I wonder if they use the VTT as input and then have another algorithm for the highlighting?

Oh, you're right. They _do_ highlight word-by-word.

jamescridland (approved):

PR #689, opened 2025-02-14
Fix <itunes:image> tags in example.xml

Pigamer37:
The image URL's were inside the tag content instead of the standard way, in the `href` attribute.

In [Apple's example feed](https://help.apple.com/itc/podcasts_connect/#/itcbaf351599) (and everywhere else I've looked) it's written the way I did on this PR.


Marzal:
LGTM

PR #711, opened 2025-06-18
Update podroll.md

jamescridland:
Adding more detail about expected items from the remoteItem spec, and some additional detail.


jamescridland:
(My thinking in adding both `feedUrl` *and* `title` into the recommended spec is that it then allows any databaseless podcast app to display podrolls without any lookups.)

daveajones:
Thanks @jamescridland !

PR #721, opened 2025-07-04
Rename purpose token "cover" to "banner"

nathangathright:
To avoid confusion with terms like [Show Cover](https://podcasters.apple.com/support/5514-show-cover-template) used by Apple for show artwork, let’s avoid the term altogether.


daveajones:
Thanks!!

PR #727, opened 2025-07-25
Remove podcast:images from examples

Marzal:
Clean up so it's not recomendad by accident.


Marzal:
Conflict resolved. I did forget about this PR.

CC @daveajones

PR #729, opened 2025-08-17
Add `poster` known purpose token

johnspurlock:


theDanielJLewis:
Why add `poster` when we can use `artwork` with `aspect-ratio="16/9"?

johnspurlock:
An item might have multiple artwork selections, and the alt enclosure video aspect ratio is not known (I'm already seeing portrait video in some feeds).  There needs to be a way for good publishers to both provide multiple aspect ratios for an item if available, and also specify _exactly one_ as the video poster.

As I understand the spec, there can be multiple tokens, right?  So if the video is 16/9, the associated `podcast:image` could have `... purpose="poster artwork" aspect-ratio="16/9" ...`

PR #766, opened 2026-03-31
chore: Improve clarity and fix typos

saerdnaer:
Rephrase sentences for clarity and correct typos in the recommendations document.


developerzeke:
> Putting detailed episode metadata into RSS feeds while keeping the full archive of all old episodes, becomes a challenge

I don't think this needs a comma

daveajones:
> > Putting detailed episode metadata into RSS feeds while keeping the full archive of all old episodes, becomes a challenge
> 
> I don't think this needs a comma

Agree.  I'll make an edit.

PR #767, opened 2026-04-11
Enhance documentation for valueRecipient tag

ChadFarrow:
Added detailed information about Lightning Address recipients and examples for using them alongside node recipients.

The example I point people to for using lnaddresses in RSS feeds at podcasting2.org is missing an example of a lnaddress in it so I always have to copy and paste an example from my feeds.

https://podcasting2.org/docs/podcast-namespace/tags/value

### DIFF
--- a/docs/tags/value-recipient.md
+++ b/docs/tags/value-recipient.md
@@ -1,5 +1,4 @@
 # Value Recipient
-
 `<podcast:valueRecipient>`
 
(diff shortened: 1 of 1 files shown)


ChadFarrow:
Screenshot 2026-04-11 at 18-30-02

This is an example from my test feed that I send people.

jamescridland:
This looks very good, thank you. I would merge if I could!

Issue #90, opened 2020-10-23
Encourage "updated" in HTTP header


theDanielJLewis:
There's a lot of unnecessary bandwidth in podcasting. Apps download the entire RSS feed every time to check if there's something new.

I propose that we set a standard for an HTTP header to reflect the last time the feed was updated. With this, smart apps could simply check the headers and if the updated date is the same as the last time the app checked, then don't download the feed again.

I think Angelo Mandato, from Blubrry, has done some work on this kind of thing before, and I think they even implemented it into Blubrry-hosted feeds or Podcast Mirror.

jamescridland:
This is a really good idea - but it's already implemented in HTTP, so we don't need to worry.

The `last-modified` header, below (from https://podnews.net/rss) does exactly the job we need.
I notice I'm not setting an eTag on this, too - which is the other way of doing it.

![Screen Shot 2020-10-23 at 11 52 20 am](https://user-images.githubusercontent.com/231941/96947329-3d850380-1526-11eb-8701-2cd7ed33ee0a.png)

theDanielJLewis:
Yes! That's exactly the tag I was thinking of.

I know it's not RSS, but I think we should somehow make this part of the spec and encourage developers to read and properly set this.

A potential problem some publishing tools might have is that they update `last-modified` whenever the feed is recached, even if nothing has changed.

daveajones:
Maybe a "best practices for hosting" section as just a companion to the spec.  I could see it being useful to some people that are self-hosting and may not be aware of these things.  Would you be willing to write up a couple of paragraphs with some bullet points on this Daniel?  If so, I'll drop it in somewhere.

dellagustin:
It would be great if this header could be trusted, of course, the best practice should mention that hosting should support the http HEAD method, so that the body (which is the heavy part) would not be downloaded.

> The HTTP HEAD method requests the headers that would be returned if the HEAD request's URL was instead requested with the HTTP GET method.

from: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD

A standard header that says "if possible only give me items published as of ..." would be nice too. Of course that would only be effective if the feed can be dynamically assembled, static file hosting would not be able to implement this.

cio-blubrry:
[...]

PR #779, opened 2026-07-08
docs: add HTTP conditional-request guidance to other-recommendations.md

conorbronsdon:
Addresses #90: Marzal's February comment asked for a PR documenting HTTP conditional requests in docs/other-recommendations.md.

The new section covers:
- Last-Modified and ETag response headers
- Honoring If-Modified-Since / If-None-Match with a 304 Not Modified response
- Why it matters at aggregator scale: Podcast Index polls millions of feeds on a schedule, and Blubrry's numbers in the issue thread show the same mechanism working at ~1.5M feeds
- Two practical notes from the thread: don't bump Last-Modified on re-cache when content hasn't changed, and confirm CDNs/hosts actually pass these headers through instead of stripping them

Follows the doc's existing heading and example style (## section, #### Examples, fenced code block).


Marzal:
Thanks, fine by me. I just would change a couple of things. Let see what others think.

CC @theDanielJLewis @daveajones

conorbronsdon:
Sounds good - happy to adjust anything needed just lmk

PR #782, opened 2026-07-28
Add Vibral service slug

HarpElle:
## Summary

Adds `vibral` to the Podcast Namespace service-slug list.

## Service contact and public-ingestion policy

- Service: Vibral
- Contact: info@vibralapp.com
- Public surfaces: Discover, public search, and trending
- Direct listener-entered RSS remains outside public directory ingestion.
- Publisher correction target: two business days.

Vibral will honor `podcast:block` for its registered slug on the public surfaces above, including the explicit `no` override defined by the Namespace specification.


Marzal:
Hi I assume, this PR is a draft because there are more incoming podcasting 2.0 features not yet implemented as
https://vibralapp.com/open-podcasting/tags/people

PR #783, opened 2026-08-01
Add Sovereigncast to element support list

Altbierjupp:
Adds Sovereigncast to the Podcasting 2.0 element support lists for Transcript, Chapters, Person, and Value.


Marzal:
Please use links that reference the tags

